PR #12, opened 2023-10-10
feat: connect wallets

ditoglez:
Currently adds possibility to connect with Metamask.


apps/idos-data-dashboard/src/routes/dashboard/attributes/index.tsx
```diff
@@ -50,6 +50,7 @@ export function Component() {
 
       if (previousAttributes) {
         const newAttribute: Attribute = {
+          id: crypto.randomUUID(),
```
juliosantos, line 53: I like this a lot — let's remove our uuid dependency everywhere 🙏

juliosantos (approved):

PR #18, opened 2023-10-19
Cleaner dapp example, demo reset, prettier UI

juliosantos:


ditoglez (approved):

apps/idos-data-dashboard/src/routes/dashboard/credentials/mutations/index.ts
```diff
@@ -42,7 +42,7 @@ export const useRemoveCredential = createMutation({
 export const useShareCredential = createMutation({
   mutationFn: async (values: Credential & { address: string; key: string }) => {
     const { key, id, address } = values;
-    return idos.grants.create("credentials", id, address, key);
+    return idos.grants.create("credentials", id, address, 0, key);
```
juliosantos, line 45: wdyt of passing `undefined` instead, and letting `NearGrants#create` do its thing?

ditoglez: I think that is a good idea, Ill address it in next PR's as probably want to redo some stuff with the whole api layer. `undefined` is fine for now.

PR #30, opened 2023-11-02
One less teleport promise

pkoch:
Top of my head, not sure if right.


pkoch:
Doesn't look dumb, merging.

PR #58, opened 2023-11-10
fix: SDK credential creation

juliosantos:
@pkoch pretty sure this is why Bob couldn't decrypt the credential Alice created for him yesterday


pkoch:
Huh. That would make sense, yeah. However, we did see the right value on the DB. So, I'm still unconvinced this would fix it.

juliosantos:
You're right — I missed a ton.

packages/idos-sdk-js/src/lib/enclave-providers/iframe-enclave.js
```diff
@@ -1,5 +1,5 @@
 export class IframeEnclave {
-  hostUrl = new URL("https://foo.com:5174");
+  hostUrl = new URL(import.meta.env.VITE_IDOS_ENCLAVE_URL);
```
juliosantos, line 2: ha, thanks @ditoglez 🙏

pkoch (approved):
I see nothing dumb. I'd have to click it a bunch to give you stronger guarantees. Let me know if you'd rather have me do that. Otherwise, ship.

pkoch:
Spoke with Júlio on a call, taking this over.

PR #81, opened 2023-12-20
chore: biome as linter and formatter

ditoglez:


sapience (approved):
I didn't check any line of code. I just believed in new linter.

PR #84, opened 2024-01-08
feat: env based chains

ditoglez:


apps/idos-data-dashboard/src/core/wagmi.ts
```diff
@@ -13,7 +13,7 @@ const metadata = {
   icons: ["/idos-dashboard-logo.svg"]
 };
 
-export const chains = [mainnet];
+export const chains = [import.meta.env.DEV ? sepolia : mainnet];
```
sapience, line 16: For local environment with using local Hardhat chain it is `localhost` needed. For `Sepolia` usage it is ok to use mainnet (it works on `staging` well).
```suggestion
export const chains = [import.meta.env.DEV ? localhost : mainnet];
```

sapience (changes requested):

sapience (approved):

PR #95, opened 2024-01-18
feat(example-dapp,sdk): EVM grants

pkoch:
Things that need to get in before this:
- [ ] https://github.com/idos-network/idos-schema/pull/36

### DIFF
--- a/apps/idos-example-dapp/package.json
+++ b/apps/idos-example-dapp/package.json
@@ -14,17 +14,21 @@
   "devDependencies": {
     "typescript": "^5.0.2",
     "vite": "^4.4.5",
-    "vite-plugin-node-polyfills": "^0.12.0",
-    "vite-plugin-mkcert": "^1.16.0"
+    "vite-plugin-mkcert": "^1.16.0",
+    "vite-plugin-node-polyfills": "^0.12.0"
   },
   "dependencies": {
     "@idos-network/idos-sdk": "workspace:*",
+    "@kwilteam/kwil-js": "0.3.0",
     "@near-wallet-selector/core": "^8.7.0",
     "@near-wallet-selector/here-wallet": "^8.7.0",
     "@near-wallet-selector/meteor-wallet": "^8.7.0",
     "@near-wallet-selector/modal-ui-js": "^8.7.0",
     "@near-wallet-selector/my-near-wallet": "^8.7.0",
     "@near-wallet-selector/nightly": "^8.7.0",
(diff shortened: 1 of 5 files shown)


packages/idos-sdk-js/src/lib/enclave.ts
```diff
@@ -43,7 +43,10 @@ export class Enclave {
     const { encryptionPublicKey, humanId, signerAddress, signerPublicKey } =
       await this.provider.load();
 
-    this.idOS.store.set("encryption-public-key", encryptionPublicKey);
+    this.idOS.store.set(
+      "encryption-public-key",
+      encryptionPublicKey && Base64Codec.encode(encryptionPublicKey)
+    );
```
pkoch, line 49: I don't see us reading this, but I saw it being wrongly serialized without being in base64. So, I fixed it.

Unsure if perhaps a better idea would be to completely yank this (and other values we write but don't read) out.

packages/idos-sdk-js/src/lib/enclave.ts
```diff
@@ -58,7 +61,10 @@ export class Enclave {
     if (!humanId) throw new Error("Could not initialize user.");
 
     this.encryptionPublicKey = await this.provider.init(humanId, signerAddress, signerPublicKey);
-    this.idOS.store.set("encryption-public-key", this.encryptionPublicKey);
+    this.idOS.store.set(
+      "encryption-public-key",
+      this.encryptionPublicKey && Base64Codec.encode(this.encryptionPublicKey)
+    );
```
pkoch, line 67: Ditto yank instead.

packages/idos-sdk-js/src/lib/enclave.ts
pkoch: Saw this working out this change. But I think we should tidy this up. Resolving.

packages/idos-sdk-js/src/lib/enclave.ts
pkoch: https://github.com/idos-network/idos-sdk-js/pull/95#discussion_r1464140597

pkoch:
Also reviewed this with Júlio on a sync call. Ready to go in.

pkoch:
> https://github.com/idos-network/idos-schema/pull/36

Derp. Forgot about this. Not too worrying. The demo will be a bit derpy for a while, but I expect to ship that soon.

PR #99, opened 2024-01-22
feat(sdk): extract kwil defaults

pkoch:
Just because I want to be able to use them in the example dApp's fake backend.


pkoch:
Things still seem to work. Confident enough to turn this into a show PR.

@ditoglez Let's iterate later if you're less than happy about this.

PR #109, opened 2024-02-09
fix: do not initialise enclave unless needed in `create`

ditoglez:


pkoch:
Lokks sane! :ok_hand:

PR #119, opened 2024-03-12
feat: enclave auth method selector

ditoglez:


apps/idos-enclave/src/dialog.js
```diff
@@ -62,28 +62,19 @@ class Dialog {
...
-        } catch (e) {
-          this.respondToEnclave({ error: e.toString() });
-        }
-      });
-    });
+    try {
+      if (type === "password") {
+        const { password } = await this.getOrCreatePasswordCredential();
+        this.respondToEnclave({ result: { password } });
+        resolve();
```
pkoch, line 69:
```suggestion
```

apps/idos-enclave/src/dialog.js
```diff
@@ -62,28 +62,19 @@ class Dialog {
...
+      } else if (type === "webauthn") {
+        const { password, credentialId } = await this.getOrCreateWebAuthnCredential();
+        this.respondToEnclave({ result: { password, credentialId } });
+        resolve();
```
pkoch, line 73:
```suggestion
```

packages/idos-sdk-js/src/lib/enclave.ts
```diff
@@ -43,7 +43,10 @@ export class Enclave {
     const { encryptionPublicKey, humanId, signerAddress, signerPublicKey } =
       await this.provider.load();
 
-    this.idOS.store.set("encryption-public-key", encryptionPublicKey);
+    if (encryptionPublicKey) {
+      this.idOS.store.set("encryption-public-key", Base64Codec.encode(encryptionPublicKey as any));
+    }
```
pkoch, line 48: Doesn't this mean we're double encoding? :thinking:

packages/idos-store/index.ts
```diff
@@ -70,8 +71,10 @@ export class Store {
     const keysInLocalStorage = Object.keys(window.localStorage);
 
     for (let key of [...new Set(keysInCookies.concat(keysInLocalStorage))]) {
-      key = key.replace(this.keyPrefix, "");
-      this.set(key, this.get(key));
+      if (key) {
+        key = key.replace(this.keyPrefix, "");
+        this.set(key, this.get(key));
+      }
```
pkoch, line 77:
```suggestion
      if (!key)  continue;
      key = key.replace(this.keyPrefix, "");
      this.set(key, this.get(key));
```

apps/idos-enclave/src/enclave.js
```diff
@@ -61,18 +72,103 @@ export class Enclave {
...
+        } else {
+          this.store.set("preferred-auth-method", "password");
+        }
+        this.store.set("password", password, duration);
+        resolve();
+      })
+    );
+  }
+
+  ensurePassword_(usePasskeys) {
```
pkoch, line 137: I don't think this is used anywhere? :thinking:

pkoch (approved):
I only have a small nit, and I keep having the same fears of double-encoding.

Since you've confirmed that you can't repro any double-encodings, I don't see any reason to stop this PR.

PR #131, opened 2024-04-10
feat: use kwil node in optimistic updates

ditoglez:


juliosantos:
Let's talk about this before merging please

PR #137, opened 2024-04-11
feat: better chainId discovery

pkoch:
- Don't use the `KwilWrapper.defaults.kwilProvider` to discover `chainId` 
- Avoid data races and stand-in values
- Use `.env` variables for the Dashboard initialisation

This already bit a user, so let's make sure we fix it.


pkoch:
I'm having some CORS problem locally, and I can't try this out.

PR #143, opened 2024-04-14
feat: add idOSGrantee API

pkoch:
Looking for critique.

Things I don't like:
- idOSBoxer is begging to be extracted. Very happy to do it.
- `grantee` address part of `idOSGrantee`. This is kind meh, since the class is couples encryption and data-fetching. Given that most users of this class will care about only their dApp network, I'm ok keeping as-is.
- `kwilNep413Signer` is also asking to be extracted, alongside with all the NEP-413 shenanigans we have spread around.
- I've extracted `KwilWrapper.init` to where it belongs. Feels like scope creep for this PR, happy to extract it and just get that in.

Known problems:
- Kwil's kgw cookie is always shared when talking to the nodes. This means a browser can't use two different signers today. I'm already in conversation with them about the issue, but I'd reckon it's a bit too low prio for them, since that's going to be a non-common use case. I think we need to bite the bullet and have a backend (unless they happen to solve the problem in time).


packages/idos-sdk-js/src/lib/idos-grantee.ts
```diff
@@ -0,0 +1,219 @@
...
+
+export type WalletType = "EVM" | "NEAR";
+
+const kwilNep413Signer =
+  (keyPair: nearAPI.KeyPair) =>
+  async (messageBytes: Uint8Array): Promise<Uint8Array> => {
+    const message = Utf8Codec.decode(messageBytes);
+    const nonceLength = 32;
+    const nonce = crypto.getRandomValues(new Uint8Array(nonceLength));
+    const recipient = "idos-example-dapp";
```
pkoch, line 23: Unsure if we want this.

PR #147, opened 2024-04-15
fix: log the kwil client out before using a new signer

pkoch:
So people don't have to manually clean HttpOnly cookies.


pkoch:
Tried it out locally using the prod nodes. Self-shipping.

PR #157, opened 2024-04-18
experimental: signer on has profile

pkoch:
To unbreak things that shouldn't be broken.


pkoch:
No need, it was an accidentally changes session secret on kgw.

PR #162, opened 2024-04-21
feat: delete multiple wallets

ditoglez:
Handles the deletion of multiple Near wallets. Groups Near wallets with the same address to look like one.


pkoch (approved):
Code looks right, and I saw this working as intended on the preview branch. :rocket:

PR #163, opened 2024-04-22
feat: use crypto implementation based on environment

ditoglez:


packages/idos-sdk-js/src/lib/idos-grantee.ts
```diff
@@ -15,6 +14,13 @@ import { implicitAddressFromPublicKey } from "./utils";
 
 export type WalletType = "EVM" | "NEAR";
 
+const crypto =
+  typeof window === "undefined" ? (await import("node:crypto")).default : window.crypto;
+
+export const getRandomValues = () => {
+  return crypto.getRandomValues(new Uint8Array(32));
+};
```
pkoch, line 22: Do we need this function at all? We're currently anchored to `crypto`.

```suggestion
```

pkoch (approved):
Gave the preview demo dapp to read creds from idOS (which uses the changed import) and it still works nice! :tada: 

Ready for prod! :rocket:

PR #164, opened 2024-04-22
chore: bump TS and vite everywhere

pkoch:
Just to keep pace.


pkoch:
This didn't work, and I need to move on to other tasks. Closing.

PR #167, opened 2024-04-22
chore: just consider crypto exists

pkoch:
I think we don't need to import anything.

```bash
$ node
Welcome to Node.js v18.19.0.
Type ".help" for more information.
> crypto.getRandomValues
[Function: getRandomValues]
```

And, it's in your window too:
image


pkoch:
Clicked around in the example dApp, everything (client- and server-side) seem to work.

PR #171, opened 2024-04-22
feat: dAG methods

pkoch:


pkoch:
No, it's lined up: https://github.com/idos-network/idos-schema/blob/b0df56003aaa6eb68c66fe10d023b1040bd9d2dd/schema.production.kf#L4

Issue #173, opened 2024-04-23
Rename all `init` methods to be convenience async builders


pkoch:
This will make it easier to understand how to navigate the codebase, since it establishes a clear pattern of where to put the code that we use to build more complicated objects.

They all should:
- Be static
- Be async
- Return a well constructed, and ready to use, self.

If a method doesn't follow this convention, it shouldn't be named `init`.

PR #175, opened 2024-04-23
chore: rename init to initialize

pkoch:
Show PR.

More work for https://github.com/idos-network/idos-sdk-js/issues/173


pkoch:
I was able to connect an EVM wallet on https://idos-data-dashboard-git-chore-rename-init-to-f95ed6-fractal-id.vercel.app/. That gives me enough confidence to self-ship this.

PR #176, opened 2024-04-23
chore: add type annotations to init functions

pkoch:
Show PR.

More work for https://github.com/idos-network/idos-sdk-js/issues/173


pkoch:
This is a type-only change on idos-sdk-js. Seeing it being successfully build gives me enough confidence to ship.

PR #178, opened 2024-04-23
chore: simpler constructor

pkoch:
Show PR, following in the spirit of https://github.com/idos-network/idos-sdk-js/issues/173.

The main purpose of this refactor is to be able to control all object creation, instead of the constructor involving execution. The intent is to make code more testable, as I can just pass in a mock instead of a `WebKwill`.


pkoch:
Was able to use Kwil (list credentials) on https://idos-example-dapp-git-chore-simpler-constructor-fractal-id.vercel.app/. Self-shipping.

PR #183, opened 2024-04-23
chore: remove `setHumanId` optimization

pkoch:
Show PR.

This used to be a useful shortcut to discover `human_id` when we required a signature on every read call to Kwil. Now that we're using kgw's cookie auth, we always need a signature, and we re-use that one for all reads.

Removing this buys having less dependencies (and therefore easier-to-test things).

### DIFF
--- a/packages/idos-sdk-js/src/lib/auth.ts
+++ b/packages/idos-sdk-js/src/lib/auth.ts
@@ -180,13 +180,6 @@ export class Auth {
     return args;
   }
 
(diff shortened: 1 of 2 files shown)


pkoch:
Saw listing and reading grants work for both EVM and NEAR on https://idos-example-dapp-git-chore-remove-sethumanid-d8c0dd-fractal-id.vercel.app/. Self-shipping.

PR #184, opened 2024-04-23
chore: remove unused arguments

pkoch:
Show PR.

Nobody is passing these arguments in. Let's just remove them.


pkoch:
I was able to clear the enclave, be asked for the password, and decrypt an existing credential on https://idos-example-dapp-git-chore-remove-unused-arguments-fractal-id.vercel.app/. Self-shipping.

PR #185, opened 2024-04-23
chore: correct StoredData

pkoch:
Show PR.

Confirmed types on localhost by adding a debugger right after `await this.provider.load()`.


pkoch:
Was able to reset the enclave and decrypt an already existing credential on https://idos-example-dapp-git-chore-correct-storeddata-fractal-id.vercel.app/. Self-shipping.

PR #186, opened 2024-04-23
chore: only logout of kwil when signer changes

pkoch:
Wide Show PR, just because it breaks current expectations.


pkoch:
Saw this both remember the cookie (since the enclave re-populates `idOS.store`) and logging me out when I changed signer on https://idos-example-dapp-git-chore-only-logout-of-kw-887675-fractal-id.vercel.app/.

Self-shipping.

PR #189, opened 2024-04-25
feat: expose idOSGrantee.createBySignature for EVM

pkoch:
Ask @ditoglez, Show @Strnadj.

I know we're getting a bit ahead of ourselves here, but @Strnadj would already benefit from having only the EVM part of this. So, I took the draft we did together, tidied up a little bit, and I'm proposing we merge it to unblock @Strnadj.

@ditoglez: Does this look passable to you? Have you perhaps iterated something on your side that I'm not aware of?


pkoch:
> Should we also expose the revoke method?

Ah, yes, good call!

PR #197, opened 2024-05-02
feat: handle expiration dates in idos store

ditoglez:


packages/idos-store/index.ts
```diff
@@ -6,6 +6,18 @@ interface PipeCodecArgs<T> {
 export class Store {
   keyPrefix = "idOS-";
 
+  constructor() {
+    for (const key in localStorage) {
+      if (key.startsWith(this.keyPrefix)) {
+        const expires = localStorage.getItem(`${key}-expires`);
+        if (expires) {
+          this.reset();
```
pkoch, line 14: Feels a bit too aggressive. I'd remove only this key.

packages/idos-store/index.ts
```diff
@@ -6,6 +6,18 @@ interface PipeCodecArgs<T> {
 export class Store {
   keyPrefix = "idOS-";
 
+  constructor() {
+    for (const key in localStorage) {
+      if (key.startsWith(this.keyPrefix)) {
```
pkoch, line 11: style nit:
```suggestion
      if (!key.startsWith(this.keyPrefix)) continue;
```

packages/idos-store/index.ts
```diff
@@ -6,6 +6,30 @@ interface PipeCodecArgs<T> {
 export class Store {
   keyPrefix = "idOS-";
 
+  constructor() {
+    for (const key in localStorage) {
+      if (key.startsWith(this.keyPrefix) && key.endsWith("-expires")) {
+        const value = localStorage.getItem(key);
+        if (value) {
```
pkoch, line 13:
```suggestion
        if (!value) continue;
```

packages/idos-store/index.ts
```diff
@@ -6,6 +6,30 @@ interface PipeCodecArgs<T> {
...
+
+          try {
+            str = JSON.parse(value);
+          } catch (error) {
+            str = "";
+          }
+
+          const expires = Date.parse(str);
+
+          if (Number.isNaN(expires) || expires < Date.now()) {
```
pkoch, line 24: > `Number.isNaN(expires)`

How is a Date a NaN? :thinking:

pkoch (commented):
Looking pretty good! :ok_hand:

packages/idos-store/index.ts
```diff
@@ -6,6 +6,30 @@ interface PipeCodecArgs<T> {
 export class Store {
   keyPrefix = "idOS-";
 
+  constructor() {
+    for (const key in localStorage) {
+      if (key.startsWith(this.keyPrefix) && key.endsWith("-expires")) {
```
pkoch, line 11:
```suggestion
      if (!key.startsWith(this.keyPrefix) || !key.endsWith("-expires")) continue;
```

packages/idos-store/index.ts
```diff
@@ -6,6 +6,30 @@ interface PipeCodecArgs<T> {
...
+        const value = localStorage.getItem(key);
+        if (value) {
+          let str;
+
+          try {
+            str = JSON.parse(value);
+          } catch (error) {
+            str = "";
```
pkoch, line 19: Do we want to do anything else if we can't decode?
```suggestion
            continue;
```

pkoch:
Talked with @ditoglez over a call. We're using the expiration mechanism to drive how long the enclave should remember your password field. However, we don't forget the private key, so the enclave will still happily work.

This is clearly not the intended functionality. The expiration time is not really per key, for but the whole enclave storage.

There is a question of supporting multiple human_ids being remembered. @juliosantos I'm going to make a product call to contain scope on this task. Namely, that we'll wipe the whole storage for now. If we want separate expiration dates keys on human_id, let's extract that to a different task.

packages/idos-store/index.ts
pkoch:
```js
> Date.parse("derp")
NaN
```

pkoch:
@ditoglez :ping_pong:

sapience (approved):

PR #207, opened 2024-05-09
feat: untangle auth deps

ditoglez:
Skipping by adding tests for `setNearSigner` method,  as discussed with @pkoch its better to manage that one with `e2e` later.

### DIFF
--- a/apps/idos-data-dashboard/src/core/idos.tsx
+++ b/apps/idos-data-dashboard/src/core/idos.tsx
@@ -97,8 +97,6 @@ export const Provider = ({ children }: PropsWithChildren) => {
       await _sdk.setSigner(signer.type, signer.value);
       const _pk = _sdk.auth.currentUser.publicKey;
 
-      console.log(_sdk.auth.currentUser);
-
       setPublicKey(_pk);
     }
 


packages/idos-sdk-js/src/lib/auth.ts
```diff
@@ -6,55 +6,60 @@ import * as Utf8Codec from "@stablelib/utf8";
...
+    if (!this.user) throw new Error("Call idOS.setSigner first.");
+    return this.user;
+  }
+
+  constructor(public readonly kwilWrapper: KwilWrapper, public readonly store: Store) {}
 
-  constructor(idOS: idOS) {
-    this.idOS = idOS;
+  static async init(kwilWrapper: KwilWrapper, store: Store) {
+    return new Auth(kwilWrapper, store);
```
pkoch, line 31: Since we're not adding any additional logic here, I think I'd rather not have this method for now.

apps/idos-data-dashboard/src/core/idos.tsx
```diff
@@ -95,7 +95,10 @@ export const Provider = ({ children }: PropsWithChildren) => {
     if (profile) {
       // @ts-expect-error
       await _sdk.setSigner(signer.type, signer.value);
-      const _pk = (await _sdk.auth.currentUser()).publicKey;
+      const _pk = _sdk.auth.currentUser.publicKey;
+
+      console.log(_sdk.auth.currentUser);
+
```
pkoch, line 101: Accidental leftover?
```suggestion

      console.log(_sdk.auth.currentUser);

```

pkoch (approved):
Looking good! A few nits to consider, but feels very shippable.

PR #213, opened 2024-05-20
chore: setup synpress and add e2e tests

ditoglez:
Setups Synpress for e2e tests, and adds a couple of tests:

### Auth:
* Sign in with a wallet (MetaMask)
* Set an EVM signer (basically sign in but with the auth flow).

### Data:
* Fetch credentials
* Fetch wallets
* Add / Delete a wallet

### Enclave:
* Decrypt a credential (enclave)

Updates the `enclave` and `dashboard` apps to use selectors for e2e tests.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -21,7 +21,7 @@ jobs:
     - uses: pnpm/action-setup@v2
     - uses: actions/setup-node@v4
       with:
-        node-version-file: './node_version'
+        node-version-file: './.node_version'
         working-directory: ./apps/idos-sdk-e2e
     - name: Install dependencies
       run: pnpm i


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,39 @@
...
+    env:
+      BASE_URL: ${{secrets.IDOS_E2E_BASE_URL}}
+      WALLET_SEED_PHRASE: ${{ secrets.IDOS_E2E_WALLET_SEED_PHRASE }}
+      WALLET_PASSWORD: ${{ secrets.IDOS_E2E_WALLET_PASSWORD }}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+    - uses: pnpm/action-setup@v2
+      with:
+        node-version: 18
```
pkoch, line 24: Can we maybe easily use a `.node-version` file for this instead?

pkoch (approved):
Great first iteration! :rocket:

PR #215, opened 2024-05-27
chore: update monorepo deps

ditoglez:
Updating dependencies. Migrates the dashboard to use the latest versions of `wagmi` `viem` and `web3modal`.


pkoch (approved):
Nothing suspicious!

PR #219, opened 2024-05-29
feat: a more sensible enclave unlocking flow

juliosantos:
* if the enclave forgot the preferred auth method but had a credential id, we were opening the popup anyway
  * now we're trying to get passkey from the iframe before opening
* if the user chooses passkey in the enclave popup and it doesn't remember the credential id, we were causing passkey creation
  * now, we assume one must exist and ask the user to choose from a passkey list instead
  * only if none is chosen AND we're in "new user" mode do we create one

**Note:** I've seen this fail occasionally during development, where at the end of the popup journey, it wouldn't close. ~~Can't reproduce it anymore, but~~ please pay extra attention to this fact, it might ring a bell during the review.

**Edit:** the above seems to be fixed; now that the Enclave uses React, waiting for the popup's native `load` event was no longer sufficient to ensure it was listening

### DIFF
--- a/apps/idos-enclave/src/lib/enclave.js
+++ b/apps/idos-enclave/src/lib/enclave.js
@@ -297,7 +297,7 @@ export class Enclave {
     const dialogURL = new URL("/dialog.html", window.location.origin);
     this.dialog = window.open(dialogURL, "idos-dialog", popupConfig);
 
-    await new Promise((resolve) => this.dialog.addEventListener("load", resolve));
+    await new Promise((resolve) => this.dialog.addEventListener("ready", resolve, { once: true }));
 
     return new Promise((resolve, reject) => {
       const { port1, port2 } = new MessageChannel();
--- a/apps/idos-enclave/src/lib/enclave.js
+++ b/apps/idos-enclave/src/lib/enclave.js
@@ -40,11 +40,15 @@ export class Enclave {
   }
 
   async keys(authMethod) {
-    if (authMethod) await this.#openDialog("auth");
-    await this.ensurePassword();
-    await this.ensureKeyPair();
-
-    return this.keyPair.publicKey;
+    try {
+      if (authMethod) await this.#openDialog("auth");
+      await this.ensurePassword();
+      await this.ensureKeyPair();
+    } catch (e) {
+      console.warn(e);
+    } finally {
+      return this.keyPair?.publicKey;
+    }
   }
 
   async authWithPassword() {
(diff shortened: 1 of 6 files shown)


Strnadj (approved):

packages/idos-sdk-js/src/lib/enclave-providers/iframe-enclave.ts
```diff
@@ -32,10 +32,11 @@ export class IframeEnclave implements EnclaveProvider {
       storage: { humanId, signerAddress, signerPublicKey },
     })) as StoredData;
 
-    if (encryptionPublicKey) return encryptionPublicKey;
+    while (!encryptionPublicKey) {
```
pkoch, line 35: This `while` triggered my wtf radar. Maybe it's worth our while to add a comment that this is meant to re-open the pop-up if the user closes it?

Come to think of it, how can the user back-off from this state now? (e.g., they decided they don't want to allow decryption anymore) I can only imagine they'll have to F5 the requesting page?

apps/idos-enclave/src/pages/methods/Passkey.tsx
```diff
@@ -64,61 +64,73 @@ export default class Passkey extends React.Component<PasskeyProps> {
...
+    const credentialRequestWithoutId = {
+      publicKey: {
+        challenge: crypto.getRandomValues(new Uint8Array(10)),
+      },
+    };
+
     if (storedCredentialId) {
-      const credentialRequest = {
+      const credentialRequestWithId = {
+        ...credentialRequestWithoutId,
```
pkoch, line 79: I'd remove this line, since we override it right after.

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -39,12 +39,15 @@ export class Enclave {
...
-    return this.keyPair.publicKey;
+  async keys() {
+    try {
+      await this.ensurePassword();
+      await this.ensureKeyPair();
+    } catch (e) {
+      console.warn(e);
+    } finally {
+      return this.keyPair?.publicKey;
+    }
```
pkoch, line 50: This finally feels weird. I'd expect something like:
```suggestion
    }
    return this.keyPair?.publicKey;
```

apps/idos-enclave/src/pages/methods/Passkey.tsx
pkoch: That is true, but this spread in specific feels useless, since there's only `publicKey` in `credentialRequestWithoutId` and we set it manually on the next line.

packages/idos-sdk-js/src/lib/enclave-providers/iframe-enclave.ts
pkoch: Hard to tell! I feel a bit too uncertain, and I'd like @juliosantos to chime in on what made them want to put this here.

pkoch (approved):
@juliosantos: Pretty sure this has everyone's blessing by now.

@ditoglez: last call for concerns!

PR #220, opened 2024-05-29
feat(example-dapp): prompt for timelock when creating a grant

ditoglez:
Prompt for a timelock whenever creating a `grant` in the `example-dapp`. It will default to `3600` if no value is provided.

This is needed mostly for e2e tests of the grants


pkoch:
I'd maybe not be as defensive, but looks pretty ok to me!

PR #225, opened 2024-05-30
feat(dashboard): prevent wallet deletion if there is only one wallet available

ditoglez:


apps/idos-data-dashboard/src/routes/dashboard/wallets/index.tsx
```diff
@@ -71,12 +71,18 @@ const WalletsList = () => {
 
   if (wallets.isSuccess) {
     const addresses = Object.keys(wallets.data);
+    console.log(addresses);
```
juliosantos, line 74: Is this deliberate?

juliosantos (approved):

PR #228, opened 2024-06-02
chore: cleanup package.json and add `publint`

ditoglez:


packages/idos-sdk-js/package.json
```diff
@@ -6,14 +6,11 @@
...
   "type": "module",
-  "main": "dist/idos-sdk.js",
-  "module": "dist/idos-sdk.umd.cjs",
-  "types": "dist/idos-sdk.d.ts",
+  "main": "src/lib/index.ts",
   "files": ["dist"],
   "exports": {
     ".": {
       "import": "./dist/idos-sdk.js",
-      "require": "./dist/idos-sdk.umd.cjs",
```
pkoch, line 16: I'm really not an expert on this, but doesn't this make the package un-require-able?

pkoch: Note I'm not aware of the current meta when it comes to importing packages. Maybe it's acceptable for folks to only publish module packages these days. :x

pkoch: Oh wow, that settles it then! :sweat_smile: 

One day, the module-only future will arrive, and this won't be a question anymore. It seems like we're getting there. <3

pkoch (approved):

PR #229, opened 2024-06-03
refactor: drop full imports of `near-api-js`

ditoglez:
Drop the full imports of `near-api-js`. Import only what is actually used by the SDK.

### DIFF
--- a/packages/idos-sdk-js/package.json
+++ b/packages/idos-sdk-js/package.json
@@ -9,9 +9,7 @@
   "main": "dist/idos-sdk.js",
   "module": "dist/idos-sdk.umd.cjs",
   "types": "dist/idos-sdk.d.ts",
-  "files": [
-    "dist"
-  ],
+  "files": ["dist"],
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


packages/idos-sdk-js/src/lib/utils.ts
```diff
@@ -1,25 +1,28 @@
...
       account_id: namedAddress,
     });
     return response.keys
-      .filter((element: object) => element.access_key.permission === "FullAccess")
-      ?.map((i) => i.public_key);
+      .filter(
+        (element: Record<string, Record<string, string>>) =>
+          element.access_key.permission === "FullAccess",
+      )
+      ?.map((i: Record<string, string>) => i.public_key);
```
pkoch, line 25: These types aren't quite right. I think we want https://github.com/near/near-api-js/blob/ed54e104fdb277eb3c91fb350f51aea41068d7a5/packages/types/src/provider/response.ts#L120, maybe?

pkoch (approved):
Only a type nit. OTT, looks shippable.

PR #232, opened 2024-06-04
feat(enclave): prevent rogue decryptions

ditoglez:


apps/idos-enclave/src/lib/enclave.js
```diff
@@ -29,22 +33,30 @@ export class Enclave {
     this.store.reset();
   }
 
-  storage(humanId, signerAddress, signerPublicKey) {
+  async storage(humanId, signerAddress, signerPublicKey) {
+    const permission = await navigator.permissions.query({
+      name: "storage-access",
+    });
```
pkoch, line 39: How well have you seen this being supported?

My first instinct here wouldn't be to try to reach in to unpartitioned state from the iframe. Instead, I'd show and communicate with the pop-up. Happy to leave this as-is if you find it less confusing.

packages/idos-store/index.ts
```diff
@@ -5,10 +5,11 @@ interface PipeCodecArgs<T> {
 
 export class Store {
   keyPrefix = "idOS-";
-
+  handler: Storage;
```
pkoch, line 8: Naming nit: I have (metaphorical) PTSD with the `handler` name. 😅 

Maybe we can call it `underlying_device` or something? It's also not super obvious to me. Happy to bike-shed in a Slack thread.

pkoch (commented):
This feels like a non-trivial change, and I'm scared of creating security breaches.

Can you enumerate the test-cases you did manually to convince yourself you're doing the right thing?

packages/idos-store/index.ts
pkoch: We settled on `device`.

pkoch:
@ditoglez Since we've taken the foot off the gas on implementing partitioned storage, should we close this PR until we pick it back up?

PR #233, opened 2024-06-06
chore: split e2e based on envs

ditoglez:


.github/workflows/playwright-playground.yml
```diff
@@ -0,0 +1,42 @@
+name: Playwright Tests (Playground)
+on:
+  push:
+    branches: [ playground ]
+  pull_request:
+    branches: [ playground ]
```
pkoch, line 6: Unsure if this is true. People will have code changes that won't be deployed on the target URLs. Right? Am I missing something?

.github/workflows/playwright-playground.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+        run:
+          working-directory: ./apps/idos-sdk-e2e
+    env:
+      BASE_URL: ${{secrets.IDOS_E2E_BASE_URL}}
+      WALLET_SEED_PHRASE: ${{ secrets.IDOS_E2E_WALLET_SEED_PHRASE }}
+      WALLET_PASSWORD: ${{ secrets.IDOS_E2E_WALLET_PASSWORD }}
+      GRANTS_TEST_BASE_URL: ${{ secrets.IDOS_E2E_GRANTS_BASE_URL }}
+      CREDENTIAL_CONTENT: ${{ secrets.IDOS_E2E_CREDENTIAL_CONTENT }}
```
pkoch, line 21: Of all of these, I'd argue that only `WALLET_SEED_PHRASE` and `WALLET_PASSWORD` are secrets. But, meh, we need some place to store those values, let's iterate on it later when it becomes cumbersome to have these as secrets.

PR #238, opened 2024-06-19
build(sdk): leverage dotenv on build steps

ditoglez:


pkoch (approved):
If that flag is enough for the `.env` file switching, ship it!

PR #239, opened 2024-06-19
feat(enclave): authorize origins for enclave operations

ditoglez:
This PR adds a list of authorized origins to the enclave, preventing rogue decryptions in browsers where storage partitioning is not a thing yet.

Tested on Chrome with Storage Partitioning disabled. If Storage Partitioning is enabled (Chrome default as of `v125`), rogue decryptions will be prevented automatically as the values of the store will be partitioned per app.

### DIFF
--- a/packages/idos-sdk-js/package.json
+++ b/packages/idos-sdk-js/package.json
@@ -7,9 +7,7 @@
   "license": "MIT",
   "type": "module",
   "main": "src/lib/index.ts",
-  "files": [
-    "dist"
-  ],
+  "files": ["dist"],
   "exports": {
     ".": {
       "import": "./dist/idos-sdk.js",


apps/idos-enclave/src/lib/enclave.js
```diff
@@ -9,6 +9,7 @@ export class Enclave {
   constructor({ parentOrigin }) {
     this.parentOrigin = parentOrigin;
     this.store = new Store();
+    this.authorizedOrigins = JSON.parse(this.store.get("enclave-authorized-origins") ?? "[]");
```
pkoch, line 12: Non-actionable: I feel a bit queasy about doing work like this in the constructor. Since we're already grabbing DOM elements a few lines below, I don't think this is making the problem worse.

However, let's keep an eye out for how hard this class is to unit-test.

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -20,6 +21,10 @@ export class Enclave {
     this.#listenToRequests();
   }
 
+  get isAuthorizedOrigin() {
+    return this.authorizedOrigins.includes(this.parentOrigin);
+  }
```
pkoch, line 26: Style: unclear if this is small enough work to make this an attribute. But, meh, 🤷, up to you.

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -31,6 +36,15 @@ export class Enclave {
...
     const storeWithCodec = this.store.pipeCodec(Base64Codec);
 
+    if (!this.isAuthorizedOrigin) {
+      return {
+        humanId: "",
+        encryptionPublicKey: "",
+        signerAddress: "",
+        signerPublicKey: "",
+      };
+    }
```
pkoch, line 46: Please no. This effectively creates a singleton value in the type's domain that all callers will have to test for. By turning this into a throw instead, we're making sure this is handled differently than a normal response more intently, and we don't pollute the type domain.

Can we make this raise instead?

pkoch (approved):

PR #246, opened 2024-06-26
chore: move unit tests to github action

ditoglez:
Removes unit tests from the precommit hook. Moves them to a dedicated Github Action.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,6 +1,7 @@
 name: Unit tests
 
 on:
+  pull_request:
   push:
     branches:
       - main


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,37 @@
+name: Unit tests
+
+on:
+  push:
+    branches:
+      - main
```
pkoch, line 6: Since unit tests (by definition), don't reach out to outside the repo, we can run this on every PR, and not just when master is pushed.
```suggestion
on:
  pull_request:
  push:
    branches:
      - main
```

pkoch (approved):
Nit'n'ship

PR #248, opened 2024-07-01
reafactor(dashboard,sdk): move an use types from sdk

ditoglez:


pkoch (approved):
I see nothing offensive! If this compiles, it seems good to go! 🚀

PR #253, opened 2024-07-16
feat(enclave,sdk): filter credentials

ditoglez:


packages/idos-sdk-js/src/lib/grants/grants.ts
```diff
@@ -170,6 +188,58 @@ class ConnectedGrants extends Grants {
...
+
+    const filteredEntriesWithContent = await Promise.all(
+      filteredEntries.map(async (credential) => {
+        const fullCredential = await this.data.get(
+          "credentials",
+          (credential as unknown as idOSCredential).id,
+          false,
+        );
+
+        if (!fullCredential) return [];
```
pkoch, line 220: Hmm, didn't think about this data race.

Maybe we wanna turn this into a `filter_map` (or whatever TS wants us to do it)?

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -254,6 +254,69 @@ export class Enclave {
...
+          const matchedValue = findValueByKey(content, key);
+          const isMatch = filterCriteria[key].includes(matchedValue);
+          if (include && !isMatch) {
+            return false;
+          }
+          if (!include && isMatch) {
+            return false;
+          }
+        }
+        return true;
```
pkoch, line 288: This feels way too complicated.

Can we use some helper methods (from lodash or some other TS helper lib)?

pkoch (commented):
On the right path, but I still wanna iterate.

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -248,12 +249,50 @@ export class Enclave {
...
+      });
+
+      return matches.filter((match) => {
+        const content = JSON.parse(match.content);
+        return filterByKeys(content, omitKeys, privateFieldFilters.omit, false);
+      });
+    };
+
+    return filterCredentials(credentials, privateFieldFilters);
+  }
```
pkoch, line 294: How about this, to remove more boilerplate?

```suggestion
  async filterCredentials(credentials, privateFieldFilters) {
    const filterByKeys = (content, filterCriteria) => {
      return every(Object.entries(filterCriteria), ([key, targetSet]) => {
        return targetSet.includes(get(content, key));
      });
    };

    const decrypted = await Promise.all(
      credentials.map(async (credential) => ({
        ...credential,
        content: Utf8Codec.decode(
          await this.decrypt(
            Base64Codec.decode(credential.content),
            Base64Codec.decode(credential.encryption_public_key),
          ),
        ),
      })),
    );

    return decrypted.filter((credential) => {
      const content = JSON.parse(credential.content);
      return filterByKeys(content, privateFieldFilters.pick) && !filterByKeys(content, privateFieldFilters.omit);
    });
  }
```

packages/idos-sdk-js/src/lib/grants/grants.ts
```diff
@@ -170,6 +188,56 @@ class ConnectedGrants extends Grants {
...
+  ): Promise<{ grant: Grant; transactionId: string }> {
+    const allEntries = (await this.data.list(tableName)) as idOSCredential[];
+
+    const filteredEntries = allEntries.filter((entry) => {
+      const keys = Object.keys(publicFields) as (keyof idOSCredential)[];
+      for (const key of keys) {
+        if (entry[key] !== publicFields[key]) return false;
+      }
+      return true;
+    });
```
pkoch, line 210: I'd maybe extract `matchCriteria` and use it here too?

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -248,12 +249,50 @@ export class Enclave {
...
+        return filterByKeys(content, positiveMatchesKeys, privateFieldFilters.pick, true);
+      });
+
+      return matches.filter((match) => {
+        const content = JSON.parse(match.content);
+        return filterByKeys(content, omitKeys, privateFieldFilters.omit, false);
+      });
+    };
+
+    return filterCredentials(credentials, privateFieldFilters);
```
pkoch, line 293: We should return just the `id`, no?

pkoch (commented):
:ping_pong:

pkoch (approved):
Code looks ok for a first approach! 👍

PR #260, opened 2024-07-25
feat(sdk): remove root enclave `container` initialisation option in favor of `enclaveOptions`

ditoglez:


packages/idos-sdk-js/src/lib/enclave-providers/iframe-enclave.ts
```diff
@@ -132,7 +133,7 @@ export class IframeEnclave implements EnclaveProvider {
     this.iframe.parentElement!.classList.remove("visible");
   }
 
-  async #requestToEnclave(request: any) {
+  async #requestToEnclave(request: unknown) {
```
blockteamdeveloper, line 136: why # you have added to function?

PR #264, opened 2024-08-01
feat: backend SDK initialisation

ditoglez:


pkoch (commented):
This looks to be on the right direction, but it's hard to tell with all the duplication.

What's your plan for tackling that?

As for deprecating things, I'm not sure if I wanna keep them around. Since we're still at v0, I'd not shy away from asking people to just start using another package.

packages/idos-sdk-server-dapp/src/idOS.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const listGrants = async (args: Partial<Omit<Grant, "lockedUntil">>) =>
+    grantee.grants?.list(args);
+  const fetchSharedCredential = async (dataId: string) =>
+    grantee.fetchSharedCredentialFromIdos(dataId);
+
+  return {
+    listGrants,
+    fetchSharedCredential,
+  };
```
pkoch, line 47: Unsure if I wanna go down this route. Let's noodle on this when you're back.

packages/idos-sdk-server-dapp/src/idOS-grantee.ts
```diff
@@ -8,16 +8,17 @@ import * as BorshCodec from "borsh";
...
+import {
+  EvmGrants,
+  type EvmGrantsOptions,
+  type NearGrantsOptions,
+} from "../../idos-sdk-js/src/lib/grants";
+import type { GrantChild } from "../../idos-sdk-js/src/lib/grants/grant-child.ts";
+import type Grant from "../../idos-sdk-js/src/lib/grants/grant.ts";
+import { KwilWrapper } from "../../idos-sdk-js/src/lib/kwil-wrapper.ts";
+import { implicitAddressFromPublicKey } from "../../idos-sdk-js/src/lib/utils.ts";
+import { assertNever } from "../../idos-sdk-js/src/types.ts";
```
pkoch, line 21: Can we get away with importing from the package base, instead of being married to a FS structure?

pkoch: Hmm, I see your point. Yeah, until we settle the API a bit more, there's not much to gain by having that hard boundary.

Dropping.

apps/idos-backend-demo-app/src/index.tsx
```diff
@@ -0,0 +1,52 @@
+import { serve } from "@hono/node-server";
+import type { Grant } from "@idos-network/idos-sdk";
```
pkoch, line 2: Only for awareness: this is begging to be re-exported on `idos-sdk-server-dapp`

packages/idos-sdk-js/src/lib/grants/grants.ts
```diff
@@ -74,7 +74,7 @@ export class Grants {
     accountId?: string;
     publicKey?: string;
   }): Promise<ConnectedGrants> {
-    let child;
+    let child: EvmGrants | NearGrants | never;
```
pkoch, line 77: nit: maybe we can narrow the type a bit?
```suggestion
    let child: EvmGrants | NearGrants;
```

packages/idos-sdk-server-dapp/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+
+### Listing grants.
+
+```js
+import {sdk} from './idOS'
+
+const grants = await sdk.listGrants({
+  owner: ""0x0000000000000000000000000000000000000000"
```
pkoch, line 29:
```suggestion
  owner: "0x0000000000000000000000000000000000000000"
```

pkoch (approved):
Looking good for a first iteration.

I think we'll inevitably end up adding support for multiple chains, but we can leave that for the first client that has that use case.

PR #265, opened 2024-08-21
feat: handle password errors with the SDK and the enclave

ditoglez:


apps/idos-enclave/src/pages/App.tsx
```diff
@@ -139,6 +142,7 @@ export function App({ store, enclave }: AppProps) {
     onError,
     onSuccess,
     mode,
+    currentUserPublicKey,
```
pkoch, line 145: We're passing this to the `Passkey` component, but we're not using it there. That smells fishy.

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -54,8 +54,8 @@ export class Enclave {
     };
   }
 
-  async keys() {
-    await this.ensurePassword();
+  async keys(currentUserPublicKey) {
```
pkoch, line 57: Careful! You missed the new argument when calling this on `async decrypt(fullMessage, senderPublicKey)`.

apps/idos-enclave/src/pages/App.tsx
```diff
@@ -64,6 +65,8 @@ export function App({ store, enclave }: AppProps) {
   const messageReceiver = useCallback((event: MessageEvent<EventData>) => {
     if (event.source !== enclave) return;
 
+    setCurrentUserPublicKey(event.data.message.currentUserPublicKey);
```
pkoch, line 68: I think we should move this to be under `case "auth"`, since that's the only message that gets this value.

apps/idos-enclave/src/pages/methods/Password.tsx
```diff
@@ -1,43 +1,66 @@
+import { decode, encode } from "@stablelib/base64";
```
pkoch, line 1: So, this violates the current importing convention, which would dictate this to be:
```suggestion
import * as Base64Codec from "@stablelib/base64";
```

Did you decide to move away from that for tree-shaking purposes? Maybe something else?

packages/idos-sdk-js/src/lib/data.ts
```diff
@@ -204,8 +205,6 @@ export class Data {
 
       const record = records.find((r: { id: string }) => r.id === recordId);
 
-      if (!record) return null;
```
pkoch, line 207: Suspicious removal. Was this on purpose?

packages/idos-sdk-js/src/lib/enclave-providers/metamask-snap-enclave.ts
```diff
@@ -93,4 +93,8 @@ export class MetaMaskSnapEnclave implements EnclaveProvider {
 
     return Uint8Array.from(Object.values(decrypted));
   }
+
+  async comparePublicKeys(encryptionPublicKey: string, idOSPublicKey: string) {
+    return this.invokeSnap("comparePublicKeys", { encryptionPublicKey, idOSPublicKey });
+  }
```
pkoch, line 99: I'd maybe abstain from adding this method at all until we start using it.

But now I'm curious: what prompted you to add it? :thinking: Was is just to have _something_ in the snap enclave, or is there maybe something I'm not seeing?

apps/idos-enclave/src/pages/methods/Password.tsx
pkoch: Yeah, these bare names really threw me off. :x

Can we follow up on this (in a different PR) to homogenize the import style? Zero rush, but I think it helps a lot.

apps/idos-enclave/src/pages/App.tsx
pkoch: I'd argue that's us circumventing the type system. I'd rather have a `MethodProps<{ password: string; duration: number }> & {expectedUserEncryptionPublicKey: string}` in the `Password` component.

Opposition?

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -176,7 +178,7 @@ export class Enclave {
   }
 
   async decrypt(fullMessage, senderPublicKey) {
-    if (!this.keyPair) await this.keys();
+    if (!this.keyPair) await this.keys(senderPublicKey);
```
pkoch, line 181: This is not true. :x You're telling the code the user should have the same pub key as the encrypter, which is false.

packages/idos-sdk-js/.env
```diff
@@ -1,4 +1,4 @@
-VITE_IDOS_ENCLAVE_URL=https://enclave.idos.network
+VITE_IDOS_ENCLAVE_URL=https://localhost:5174
```
pkoch, line 1: Accidental inclusion?

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -309,7 +313,7 @@ export class Enclave {
           keys: () => [],
           reset: () => [],
           configure: () => [mode, theme],
-          storage: () => [humanId, signerAddress, signerPublicKey],
+          storage: () => [humanId, signerAddress, signerPublicKey, expectedUserEncryptionPublicKey],
```
juliosantos, line 316: Since I have you here, please see if we can get rid of `signerAddress` and `signerPublicKey` which don't seem to be in use anymore — if we can, please do it on a separate PR.

apps/idos-enclave/src/pages/methods/Password.tsx
```diff
@@ -1,43 +1,70 @@
...
-}: MethodProps<{ password: string; duration: number }>) {
+  encryptionPublicKey,
+}: MethodProps<{ password: string; duration: number }> & { encryptionPublicKey: string }) {
   const [password, setPassword] = useState("");
   const [duration, setDuration] = useState(7);
   const passwordInput = useRef<{ focus: () => void }>(null);
+  const [error, setError] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  async function derivePublicKeyFromPassword(password: string) {
```
juliosantos, line 23: What do you think of, instead of duplicating code and imports over here, sending a message back to the iframe enclave with the password string and deriving the key over there?

apps/idos-enclave/src/pages/methods/Password.tsx
```diff
@@ -1,43 +1,70 @@
...
   const onSubmit = useCallback(
-    (e: Event) => {
+    async (e: Event) => {
       e.preventDefault();
       e.stopPropagation();
-
+      setIsLoading(true);
+
+      const derivedPK = await derivePublicKeyFromPassword(password);
+      if (derivedPK !== encryptionPublicKey) {
```
juliosantos, line 41: Please make sure this comparison is only done when a user has a `current_public_key`, otherwise it will fail for new users. Same thing regarding existing users for whom `current_public_key` will be empty, unless Alex backfilled the database for everyone.

sapience: All users must have `current_public_key`. 
https://github.com/idos-network/idos-schema/blob/master/schema.production.kf#L29
```sql
table humans {
    id uuid primary,
    current_public_key text notnull,
    inserter text notnull
}
```

juliosantos: @sapience can you shed a little light on this backfill please?

juliosantos (approved):
Lgtm, but before merging please exercise this across not just the dashboard but the example dapp and Fractal ID

PR #266, opened 2024-08-27
feat(enclave):  refactor things a little bit on passkeys

ditoglez:


juliosantos (changes requested):
Since this is a refactor, I think we should take the opportunity here to remove the functionality related to password credentials. That's left over from past experiments that we decided not to enable, and removing it will make these files a lot simpler.

PR #268, opened 2024-09-09
feat(enclave): password / key recovery functionality

ditoglez:
- [x] Recovery functionality on the Secure Enclave.
- [x] Dashboard "Settings" route that will trigger the backup functionality.
- [x] Store Password / Secret on Google Drive.
- [x] Store Password / Secret within the idOS.
- [x] Overall UX  and flow improvements

### DIFF
--- a/packages/idos-sdk-js/src/lib/idos.ts
+++ b/packages/idos-sdk-js/src/lib/idos.ts
@@ -176,7 +176,6 @@ export class idOS {
     }
 
     if (attributeToCreate.length) {
-      console.log({ attributeToCreate });
       await this.data.createMultiple("attributes", attributeToCreate, true);
     }
   }


apps/idos-enclave/src/pages/recovery/AccountRecovery.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  if (reveal.value) {
+    return <PasswordAndSecretReveal {...{ password, secret }} onCancel={handleReveal} />;
+  }
+
+  return (
+    <div class="flex flex-col gap-5">
+      <Heading>Recover your password or secret key</Heading>
+      <Button onClick={handleReveal}>Reveal password / secret</Button>
+      <Button onClick={handleDownload}>Download password / secret</Button>
```
juliosantos, line 127:
```suggestion
      <Button onClick={handleReveal}>Reveal password / secret key</Button>
      <Button onClick={handleDownload}>Download password / secret key</Button>
```

juliosantos (changes requested):
Aside from the 2 nits I left, there's a big one I'd like you to address: I think "Back up" is a much better term for the concept of this PR than "Recover".

Recovery is what a user would do with the backup they made. It doesn't really apply to password (since they can just check their backup and type it), but for the passkey scenario we'll need a way for them to enter their secret key for us to store it in a newly generated passkey.

If you agree, please make sure to reflect this not just in the copy, e.g.

```tsx
-      <Heading>Recover your password or secret key</Heading>
+      <Heading>Back up your password or secret key</Heading>
```

but also in the code and filenames, e.g.

```bash
mv apps/idos-enclave/src/pages/recovery/AccountRecovery.tsx \
   apps/idos-enclave/src/pages/recovery/PasswordOrKeyBackup.tsx
```

If you don't agree, counterargue :)

juliosantos (changes requested):

PR #271, opened 2024-09-10
feat(dashboard): add settings route

ditoglez:

PR #273, opened 2024-09-16
feat(enclave): store idOS credentials in google drive

ditoglez:


juliosantos (changes requested):
The Drive stuff lgtm, I think that's the only new stuff after https://github.com/idos-network/idos-sdk-js/pull/268 and https://github.com/idos-network/idos-sdk-js/pull/271, right?

The comments I left on those two apply here too

PR #294, opened 2024-10-03
feat: optional networks

pkoch:
Made sure it works by clicking around in idos-example-dapp.

### DIFF
--- a/apps/idos-example-dapp/vite.config.ts
+++ b/apps/idos-example-dapp/vite.config.ts
@@ -66,6 +66,7 @@ export default defineConfig({
     target: "esnext",
     rollupOptions: {
       manualChunks: {
+        ethers: ["ethers"],
         "near-api-js": ["near-api-js"],
       },
     },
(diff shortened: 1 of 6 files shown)


pkoch:
This is making the example dapp hang. No good clue why:
image

pkoch:
Messed around with this for a bit, but it seems like the code splitting part on vite is not doing a great job. Namely, it seems to include not only ethers but also the code that imports it! :shrug:

Not sure what to do with this, letting it marinate here until it's a priority again.

pkoch:
It does work for near, tho.

pkoch:
For reference, it does work just fine locally.

pkoch:
@ditoglez I had to add this to `apps/idos-example-dapp/vite.config.ts`:
```typescript
export default defineConfig({
  build: {
    target: "esnext",
    rollupOptions: {
      manualChunks: {
        ethers: ["ethers"],
        "near-api-js": ["near-api-js"],
      },
    },
  },
 ...etc
});
```

I think this spells trouble for us, since we'll have to teach people how to use us with various bundlers, since only the app bundler can do code splitting.

But, this is working for now, and that's good enough for me to merge.

PR #297, opened 2024-10-03
fix: type nits

pkoch:
Just to get some easy wins in.


pkoch:
This is a type-only change. Self-shipping.

PR #298, opened 2024-10-03
feat: remove profile providers

pkoch:
To be able to remove all references in the docs.


pkoch:
Self-shipping. I've seen the preview branch work as expected.

PR #312, opened 2024-10-15
feat(issuer): allow issuers to pass a generated id to payloads

ditoglez:


packages/issuer-sdk-js/src/human.ts
```diff
@@ -42,14 +46,14 @@ async function upsertWallet(
...
   config: CreateIssuerConfig,
-  current_public_key: string,
-  wallet: Omit<UpsertWalletReqParams, "id" | "human_id">,
+  human: CreateProfileReqParams,
+  wallet: Omit<UpsertWalletReqParams, "human_id">,
 ) {
-  const human_id = crypto.randomUUID();
-  const wallet_id = crypto.randomUUID();
+  const human_id = human.id ?? crypto.randomUUID();
+  const wallet_id = wallet.id ?? crypto.randomUUID();
```
Mohammed-Mamoun98, line 54: how about another method that can also be used on createActionInput. we can call it sth like ensureId
```
const ensureId = (id?:string) => id || crypto.randomUUID() 
// use it else where
const human_id = ensureId(human.id)
 const wallet_id = ensureId(wallet.id)
const payload = { ...params, id: ensureId(params.id)};
```

Mohammed-Mamoun98 (approved):
LGTM

PR #319, opened 2024-10-18
feat(client-sdk): refactor data.ts

ditoglez:


packages/idos-sdk-js/src/lib/idos.ts
```diff
@@ -63,7 +66,13 @@ export class idOS {
...
     );
 
-    this.data = new Data(kwilWrapper, this.enclave);
+    this.data = new Data(
+      kwilWrapper,
+      this.enclave,
+      new AttributeDataLayer(this.enclave, kwilWrapper),
+      new CredentialDataLayer(this.enclave, kwilWrapper),
+      new WalletDataLayer(this.enclave, kwilWrapper),
+    );
```
pkoch, line 75: Please make the order of `kwilWrapper` and `this.enclave` consistent between `Data` and `DataLayer` children. Just trying to avoid mistakes.

packages/idos-sdk-js/src/lib/data/datalayer.ts
```diff
@@ -0,0 +1,134 @@
...
+      `edit_${this.singularize(this.tableName)}`,
+      [record],
+      description,
+      synchronous,
+    );
+
+    return record;
+  }
+
+  async share(id: string) {
```
pkoch, line 109: Not all data can be shared. Only credentials and attributes.

Let's maybe make an intermediary class for that?

packages/idos-sdk-js/src/lib/data/datalayer.ts
```diff
@@ -0,0 +1,134 @@
...
+    const records = await this.list();
+    const record = records.find((r) => r.id === id);
+    return record || null;
+  }
+
+  async getShared(id: string): Promise<T | null> {
+    const records = await this.list();
+    const record = records.find((r) => r.id === id);
+    return record || null;
+  }
```
pkoch, line 67: Ditto sharing

packages/idos-sdk-js/src/lib/data/datalayer.ts
```diff
@@ -0,0 +1,134 @@
...
+      [payload],
+      `Share a ${this.entityName} on idOS`,
+    );
+
+    return payload;
+  }
+
+  async unshare(id: string): Promise<{ id: string }> {
+    return await this.delete(id);
+  }
```
pkoch, line 133: Ditto sharing

pkoch (commented):
Feels like a step in the right direction! :+1:

However, do keep in mind you need to update the other apps and libs too! Most notably, the idos-example-dapp.

PR #326, opened 2024-10-23
Update Idos App's dependencies versions

Mohammed-Mamoun98:


apps/idos-data-dashboard/src/core/idos/idos-provider.tsx
```diff
@@ -55,6 +55,8 @@ export const Provider = ({ children }: PropsWithChildren) => {
         url: import.meta.env.VITE_IDOS_ENCLAVE_URL,
       },
     });
+    //@ts-ignore
+    window._sdk = _sdk;
```
pkoch, line 59: @ditoglez any concerns?

pkoch (approved):
I have a point that I was @ditoglez to also ship.

From my part, nothing looks dumb at all! As long as you've seen all the apps work, I think we're good!

apps/idos-data-dashboard/src/core/idos/idos-provider.tsx
Mohammed-Mamoun98: wrong push. my bad. reverting rn.

PR #349, opened 2024-11-12
remove the need for passing issuer's encryption public key at credent…

Mohammed-Mamoun98:
…ial creation

### DIFF
--- a/packages/issuer-sdk-js/src/create-issuer-config.test.ts
+++ b/packages/issuer-sdk-js/src/create-issuer-config.test.ts
@@ -1,5 +1,7 @@
 import { KwilSigner, NodeKwil } from "@kwilteam/kwil-js";
+import * as Base64Codec from "@stablelib/base64";
 import { Wallet } from "ethers";
+import nacl from "tweetnacl";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { createIssuerConfig } from "./index";
 
@@ -40,10 +42,12 @@ describe("createIssuerConfig", () => {
     const mockWallet = new Wallet(
       "dcdda6663be8dfa23d0e54a31ff6fddba2fdf8a1f0eae985c59857031e6da169",
     );
+    const mockSecretKey = nacl.randomBytes(32);
+
     const params = {
       nodeUrl: "http://mock-node-url",
       signer: mockWallet,
-      encryptionSecret: "mock-secret-key",
+      encryptionSecret: Base64Codec.encode(mockSecretKey),
     };
 
     const result = await createIssuerConfig(params);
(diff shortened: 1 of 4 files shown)


packages/issuer-sdk-js/src/create-issuer-config.test.ts
```diff
@@ -40,10 +42,12 @@ describe("createIssuerConfig", () => {
     const mockWallet = new Wallet(
       "dcdda6663be8dfa23d0e54a31ff6fddba2fdf8a1f0eae985c59857031e6da169",
     );
+    const mockSecretKey = nacl.randomBytes(32);
```
pkoch, line 45: Let's use this interface to make sure we get the right bytes:
```suggestion
    const mockSecretKey = nacl.box.keypair().secretKey;
```

packages/issuer-sdk-js/src/create-issuer-config.ts
```diff
@@ -48,8 +50,10 @@ export async function createIssuerConfig(params: CreateIssuerConfigParams) {
   });
 
   const signer = createKwilSigner(params.signer);
+  const secretKey = Base64Codec.decode(params.encryptionSecret);
+  const keyPair = nacl.box.keyPair.fromSecretKey(secretKey);
 
-  return { chainId, dbid, kwilClient, signer, encryptionSecret: params.encryptionSecret };
+  return { chainId, dbid, kwilClient, signer, keyPair };
```
pkoch, line 56: Please use names that indicate the key's usage.
```suggestion
  return { chainId, dbid, kwilClient, signer, encrypter };
```

packages/issuer-sdk-js/src/credentials.ts
```diff
@@ -4,30 +4,37 @@ import type { idOSCredential } from "../../types";
 import type { IssuerConfig } from "./create-issuer-config";
 import { createActionInput, encryptContent, ensureEntityId } from "./internal";
 
-// Base interface for credential parameters
-interface BaseCredentialParams extends Omit<idOSCredential, "id" | "original_id"> {
+/**
+ * Interface for BaseCredentialParams.
+ *
+ * `id` could be provided manually, but it is randomly generated internally.
```
pkoch, line 10:
```suggestion
 * `id` can be manually provided. If it isn't provided, it will be randomly generated.
```

packages/issuer-sdk-js/src/credentials.ts
```diff
@@ -4,30 +4,37 @@ import type { idOSCredential } from "../../types";
...
+ * `encryption_public_key` is **excluded** because it is derived internally
+ * from the issuer's secret key and should not be provided manually.
```
pkoch, line 12:
```suggestion
 * `encryption_public_key` is omitted because it is derived internally
 * from the issuer's encryption secret key.
```

PR #367, opened 2024-11-27
feat(client-sdk): allow to pass `synchronous` flag for kwil executions

ditoglez:


pkoch (commented):
Let's instead force it to be `true` on kwil wrapper. No need for more variables.

pkoch (approved):
Didn't try it out, but the change looks good. :ok_hand:

PR #379, opened 2024-12-05
fix(enclave): less recursion

pkoch:
Just trying to reduce branching.


Mohammed-Mamoun98:
already merged in another branch

PR #380, opened 2024-12-08
feat: remove extra iframe creation

Mohammed-Mamoun98:


pkoch (approved):

packages/idos-sdk-js/src/lib/enclave-providers/iframe-enclave.ts
```diff
@@ -152,11 +147,23 @@ export class IframeEnclave implements EnclaveProvider {
       this.iframe.style.setProperty(k, v);
     }
 
-    if (!container) throw new Error(`Can't find container with selector ${this.container}`);
-
+    let el: HTMLElement | null;
+    // biome-ignore lint/suspicious/noAssignInExpressions: it's on purpose
+    while ((el = document.getElementById(this.iframe.id))) {
+      console.log("reinstalling idOS iframe...");
```
pkoch, line 153: Come to think of it, I'd rather we produced less noise in the console during normal operation.

Unsure what to do here, since this might have more of an impact. Staying put for now.

PR #385, opened 2024-12-10
feat(issuer-sdk, enclave): use ephemeral encryption private keys

ditoglez:


apps/idos-enclave/src/lib/enclave.js
```diff
@@ -171,7 +171,8 @@ export class Enclave {
 
   encrypt(message, receiverPublicKey = this.keyPair.publicKey) {
     const nonce = nacl.randomBytes(nacl.box.nonceLength);
-    const encrypted = nacl.box(message, nonce, receiverPublicKey, this.keyPair.secretKey);
+    const ephemeralKeyPair = nacl.box.keyPair();
```
pkoch, line 174: Just realized this has a flaw: the information about the public key is lost, and we'll need it for sure on AGs (since it's not `this.keyPair.publicKey` anymore).

pkoch (commented):
We're missing a wire.

pkoch (approved):
Looks good on paper! Give it a spin, and we're good to go!

PR #387, opened 2024-12-12
always pass EPEK to enclave wether it remembers auth method or not

Mohammed-Mamoun98:
EPEK 👉 Expected public Encryption Key


apps/idos-enclave/src/pages/App.tsx
```diff
@@ -102,11 +102,11 @@ export function App({ store, enclave }: AppProps) {
       throw new Error(`Unexpected request from parent: ${requestData.intent}`);
 
     responsePort.current = ports[0];
+    setEncryptionUserPublicKey(event.data.message.expectedUserEncryptionPublicKey);
```
juliosantos, line 105:
```suggestion
    setEncryptionUserPublicKey(requestData.message?.expectedUserEncryptionPublicKey);
```
This seems unsafe just looking at the rest of the file, wdyt of this alt?

juliosantos: Also, how does `setEncryptionUserPublicKey` handle empty/null/undefined values?

Mohammed-Mamoun98: we can either throw an error "No Public encryption key found" OR just log and error with same message. what do u think?

pkoch: Throw. Let's make it loudly blow up.

apps/idos-enclave/src/pages/App.tsx
```diff
@@ -102,11 +102,11 @@ export function App({ store, enclave }: AppProps) {
       throw new Error(`Unexpected request from parent: ${requestData.intent}`);
 
     responsePort.current = ports[0];
+    setEncryptionUserPublicKey(requestData.message.expectedUserEncryptionPublicKey);
```
juliosantos, line 105:
```suggestion
    setEncryptionUserPublicKey(requestData.message?.expectedUserEncryptionPublicKey);
```

We're using safe navigation just a few lines below, should't we do the same here?

juliosantos (approved):

PR #388, opened 2024-12-12
chore: add `codecs` library and refactor client and issuer sdk's to use it

ditoglez:


pkoch (commented):
Not opposed at all. Let's just make sure we use the formulation that most facilitates tree shaking.

pkoch (approved):
Careful with the conflicts. Other than that, :rocket:.

PR #389, opened 2024-12-13
chore(issuer-sdk-demo): downgrade supports-color package

ditoglez:


pkoch (approved):
As long as it works! :rocket:

PR #394, opened 2024-12-17
feat(dashboard-for-dapps): credential content filtering

ditoglez:
Current PR adds the following functionality to the Dashboard for dApps:
- [x] Fetch all credentials for a grantee and decrypt them locally
- [x] Implement client-side pagination for large numbers of data
- [x] Implement client-side filtering
- [x] Prompt for the `SECRET_KEY` upon decryption if its not known


Mohammed-Mamoun98:
on credential page. and there's no stored grantee/owner secret key am showing this 
image
and clicking on setting secret key will open the secret key prompt
image
after setting it start fetching creds
image

apps/dashboard-for-dapps/src/routes/index.tsx
```diff
@@ -334,15 +216,17 @@ function CredentialDetails({
...
-                            rounded="md"
-                            loading="lazy"
-                            width="120px"
-                            height="120px"
-                            title="Click to open the image in full size"
-                          />
+                          <chakra.button className="button">
+                            <Image
+                              src={value}
+                              alt="Identification document front"
```
juliosantos, line 222: If we're hardcoding alternate text, we should probably make it more generic no? Like "Image from credential"

apps/dashboard-for-dapps/src/routes/index.tsx
```diff
@@ -58,14 +49,6 @@ export const Route = createFileRoute("/")({
...
 });
 
-function transformBase85Image(src: string) {
-  const prefix = "data:image/jpeg;base85,";
-
-  return `data:image/png;base64,${Base64Codec.encode(
-    ascii85.decode(src.substring(prefix.length)),
-  )}`;
-}
-
```
juliosantos, line 68: How does image rendering work without this? I remember picking base85 because of superior compression, but it's a format browsers don't render it natively, which is why we're converting it with this function or similar. I may be misunderstanding, but it seems that now you're calling `openImageInNewTab` with base85 content. Is this a format browsers render now, or am I missing something else?

Mohammed-Mamoun98: we used to use this method indeed. but appearantly it was refactored to openInNewTab. didn't notice that it doesn't work properly but taking a look rn

juliosantos (approved):

PR #411, opened 2024-12-30
Fix(dashboard-for-dapps): dapps dashboard fixes

Mohammed-Mamoun98:


apps/dashboard-for-dapps/src/routes/credentials.tsx
```diff
@@ -443,18 +445,23 @@ function Credentials() {
...
-    if (!deferredSearchItem) return credentials.data;
-    return matchSorter(credentials.data || [], deferredSearchItem, {
+    if (!decryptedCredentials.data) return [];
+    if (!deferredSearchItem) return decryptedCredentials.data;
+    return matchSorter(decryptedCredentials.data || [], deferredSearchItem, {
       keys: ["public_notes", "content"],
       threshold: matchSorter.rankings.CONTAINS,
     });
-  }, [deferredSearchItem, credentials.data]);
+  }, [deferredSearchItem, decryptedCredentials.data, credentialsList.data]);
```
pkoch, line 464: Why did we end up depending on `credentialsList`?

apps/dashboard-for-dapps/src/routes/index.tsx
```diff
@@ -92,14 +94,16 @@ function CredentialDetails({
...
+  const hasValidContent = !!result;
 
   const subject = Object.entries(content.credentialSubject).filter(
     ([key]) => !["emails", "wallets"].includes(key) && !key.endsWith("_file"),
   ) as [string, string][];
 
-  const emails = content.credentialSubject.emails;
-  const wallets = content.credentialSubject.wallets;
+  const emails = content.credentialSubject?.emails || [];
+  const wallets = content.credentialSubject?.wallets || [];
```
pkoch, line 106: How easy would it be to declare the types here? That'd avoid us having to specify them when we call `.map` further down.

pkoch (commented):
Spirit looks good.

apps/dashboard-for-dapps/src/routes/credentials.tsx
Mohammed-Mamoun98: you're right decrypted already has credentialsList as a dep. lemme check the effect of this change

apps/dashboard-for-dapps/src/routes/index.tsx
Mohammed-Mamoun98: I managed to type them but got multible TS errors. it's fixable but i don't think it's currently worth the time.
your thoughts?

pkoch: Let me have a quick try.

pkoch: Pushed a commit with some simple fixes.

pkoch (approved):

PR #419, opened 2025-01-07
feat(client-sdk): calculate credential content sha256 hash

ditoglez:


packages/idos-sdk-js/src/lib/data.ts
```diff
@@ -367,6 +367,18 @@ export class Data {
     });
   }
 
+  async getCredentialContentSha256Hash(credentialId: string) {
```
Mohammed-Mamoun98, line 370: isn't gonna be more effecient to pass the credential itself. since u can pass the Cred id so u do have an access to the whole credentail. that means another re-fetch for the credentail.

packages/idos-sdk-js/src/lib/data.ts
```diff
@@ -367,6 +367,18 @@ export class Data {
...
+    const credential = (await this.get("credentials", credentialId, false)) as idOSCredential;
+
+    const encoder = new TextEncoder();
+    const encodedContent = encoder.encode(credential.content);
+    const buffer = await window.crypto.subtle.digest("SHA-256", encodedContent);
```
Mohammed-Mamoun98, line 375: do u ``window`` could make an issue if used on Next apps or Node.js?

PR #550, opened 2025-02-20
feat(issuer-sdk): make near-api-js and ethers optional

Strnadj:


packages/consumer-sdk-js/src/server/index.ts
```diff
@@ -1,7 +1,3 @@
-import { Wallet } from "ethers";
-import { JsonRpcProvider } from "ethers";
-import { KeyPair } from "near-api-js";
```
pkoch, line 3: We should also declare these as optional on `packages/consumer-sdk-js/package.json`, just like on `packages/idos-sdk-js/package.json` (search for `"peerDependenciesMeta"`). No?

PR #745, opened 2025-04-03
feat: remove passkeys

### LINKED ISSUE #729: Remove passkey support
Their current implementation is finicky, and we're planning using other options for easier key management. For now, let's get rid of them.

ditoglez:
Closes #729

### DIFF
--- a/apps/idos-enclave/src/lib/enclave.js
+++ b/apps/idos-enclave/src/lib/enclave.js
@@ -42,9 +42,6 @@ export class Enclave {
   }
 
   storage(userId, expectedUserEncryptionPublicKey) {
-    // In case the user is different, we reset the store.
-    if (userId !== this.store.get("user-id")) this.reset();
-
     userId && this.store.set("user-id", userId);
     const storeWithCodec = this.store.pipeCodec(Base64Codec);
 
--- a/apps/isle/src/features/create-profile.tsx
+++ b/apps/isle/src/features/create-profile.tsx
@@ -1,7 +1,6 @@
 import { Center, Heading, Text, VStack } from "@chakra-ui/react";
 import { useState } from "react";
 
-import { WalletIcon } from "@/components/icons/wallet";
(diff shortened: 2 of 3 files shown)


apps/idos-enclave/src/app.tsx
```diff
@@ -0,0 +1,224 @@
...
+          message.value = requestData.message?.message;
+          break;
+
+        case "backupPasswordOrSecret":
+          isBackupMode.value = true;
+          backupStatus.value = requestData.message?.status;
+          break;
+      }
+
+      if (requestData.configuration.mode) mode.value = requestData.configuration.mode;
```
pkoch, line 139: Style alternative. Just pointing it out, unsure which I like more:
```typescript
mode.value = requestData.configuration.mode ?? mode.value
```

apps/idos-enclave/src/app.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  return (
+    <Layout>
+      <PasswordForm
+        {...methodProps}
+        encryptionPublicKey={encryptionPublicKey.value}
+        userId={userId.value}
+      />
+    </Layout>
+  );
```
pkoch, line 223: Don't action: this, in my eyes, begs for the same type of union style we ended up at with idOSClient.

pkoch (commented):
Great work! :clap:

apps/idos-enclave/src/lib/enclave.js
```diff
@@ -125,14 +98,6 @@ export class Enclave {
         this.authorizedOrigins = [...new Set([...this.authorizedOrigins, this.parentOrigin])];
         this.store.set("enclave-authorized-origins", JSON.stringify(this.authorizedOrigins));
 
-        if (credentialId) {
-          this.store.set("credential-id", credentialId);
```
pkoch, line 129: We should start wiping this out from the store, too. Maybe I missed the part where we do that?

PR #769, opened 2025-04-07
Fix/even better types

pkoch:
Should be a behavior-preserving PR.

Ask PR for sanity-checking.


sapience (approved):
Don't see anything criminal.
If it is tested, then ship it :ship

PR #819, opened 2025-05-01
Draft guides and appendixes

juliosantos:


pkoch (approved):
Still not complete, but it gets the point across.

PR #833, opened 2025-05-06
feat(client-sdk): add credential filtering

ditoglez:


packages/@core/src/client/index.ts
```diff
@@ -328,4 +335,35 @@ export class idOSClientLoggedIn implements Omit<Properties<idOSClientWithUserSig
...
+
+    const credentials = await this.getAllCredentials();
+    const originalCredentials = credentials.filter((cred) => !cred.original_id);
+
+    const matchingCredentials = originalCredentials.filter((cred) => {
+      const publicNotes = safeParse(cred.public_notes);
+      return acceptedIssuers?.some(
+        (issuer) =>
+          issuer.authPublicKey === cred.issuer_auth_public_key &&
+          acceptedCredentialType === publicNotes?.type,
```
pkoch, line 360: Maybe we should do the same `pick` and `omit` trick we do for private fields?

PR #846, opened 2025-05-08
fix(isle-controller): pass wagmi config at isle controller setup

Mohammed-Mamoun98:


pkoch (approved):
Shipping.

However, @ditoglez: please take a look to see if this raises any wallet switching concerns for you.

PR #875, opened 2025-05-15
chore: update packages

ditoglez:


Mohammed-Mamoun98 (commented):
Please update README related files. idos-sdk-js still has outdated Grant type

PR #878, opened 2025-05-16
fix(data-dashboard): Fix Near wallet connection

Mohammed-Mamoun98:


pkoch (commented):
I can't review the spirit of the code, I'm going to have to leave that to @ditoglez.

Mechanically, I see nothing obviously wrong.

PR #899, opened 2025-06-05
feat: add near support for passporting demo

ditoglez:


packages/@core/src/kwil-infra/create-near-wallet-kwil-signer.ts
```diff
@@ -35,6 +35,54 @@ export function looksLikeNearWallet(signer: unknown): signer is NearWallet {
...
+  if (!result || !result.signature) {
+    throw new Error("Failed to get signature from NEAR wallet");
+  }
+
+  // Create the full payload
+  const fullPayload = bytesConcat(
+    binaryWriteUint16BE(nep413BorshPayload.length),
+    nep413BorshPayload,
+    base64Decode(result.signature),
+  );
```
sapience, line 81: This is duplicate of what [signer does](https://github.com/idos-network/idos-sdk-js/pull/899/commits/ac91093724049e2f21414101399b2b7214a99d73#diff-790160c7fefa17e93714d20e1c5165e1ebcf57a22b482ab6de94b5367b45bbc1R173-R225)
We should refactor it and reuse. 
So it should be look something like
```js
export const signNearMessage = async (wallet: NearWallet, message: string): Promise<string> => {
  nep413Bytes = signer(wallet, message)
  return Buffer.from(nep413Bytes).toString("base64");
};
```

PR #979, opened 2025-07-21
feat(neobank): stateless session management

ditoglez:


Mohammed-Mamoun98 (commented):
please check `examples/pay-demo/app/providers/sessions.server.ts` it shows a reusable variable in a singltone way

PR #982, opened 2025-07-22
fix(enclave): Backup or reveal secret

Strnadj:


apps/idos-enclave/src/features/recovery/backup.tsx
```diff
@@ -187,7 +186,7 @@ export function PasswordOrSecretReveal({
   );
 }
 
-export async function PasswordOrKeyBackup({
```
Strnadj, line 190: I don't know what I was thinking about... this is a temporary fix, since in the new version on which I am working on, this will be solved in different way.

apps/idos-enclave/src/features/recovery/backup.tsx
```diff
@@ -197,24 +196,34 @@ export async function PasswordOrKeyBackup({
   onSuccess: (result: unknown) => void;
 }) {
   const reveal = useSignal(false);
-  const status = useSignal<"idle" | "pending">("idle");
+  const authMethod = useSignal<"passkey" | "password" | null>(null);
+  const password = useSignal<string | null>(null);
+
+  useEffect(() => {
+    store.get<"passkey" | "password">("preferred-auth-method").then((method) => {
+      if (method) {
```
Strnadj, line 204: horrible but temporary

PR #987, opened 2025-07-23
feat(consumer-and-isssuer): use credentials builder

Strnadj:


examples/consumer-and-issuer/src/actions/index.ts
```diff
@@ -1,99 +1,101 @@
...
+  selfieFile: Buffer,
+) => {
+  const { info, review, email, phone } = kycData;
+  const idDoc = info.idDocs[0];
 
-  const body = {
-    xsd: "http://www.w3.org/2001/XMLSchema#",
-    aux: "https://raw.githubusercontent.com/trustfractal/claim-schemas/master/aux.xml",
+  // TODO: Fix this
+  const issuerHost = "https://consumer-and-issuer-demo.playground.idos.network";
```
Strnadj, line 28: @ditoglez what should we use here?

Strnadj: because `process.env.VERCEL_URL` can't be used... it's bind to one ... but we are in demo... so maybe it's not so bad

PR #997, opened 2025-07-28
2/4 feat (utils#enclave) Store password method store in kwill

Strnadj:


apps/idos-enclave/src/features/auth/index.tsx
```diff
@@ -19,25 +22,25 @@ export default function AuthMethodChooser({
...
-    if (authMethod.value === "mpc" && allowedAuthMethods.includes("mpc")) {
+    if (currentPasswordStore.value === "mpc" && allowedEncryptionStores.includes("mpc")) {
       // There is no next step, no password or duration
-      onSuccess({ authMethod: authMethod.value });
+      onSuccess({ encryptionPasswordStore: currentPasswordStore.value });
     }
   });
 
-  if (authMethod.value === "password") {
+  if (currentPasswordStore.value === "password") {
```
sapience, line 43: The name `password` as store type may confuse developers. Because despite of type of storage we store a `password`. If we use MPC storage, than we store the password on MPC nodes. Otherwise a user by himself stores/remembers the password, so technically the store is the `user` (his head, his google drive etc.). That is why in the schema we [have](https://github.com/idos-network/idos-schema/pull/87/files#diff-939acd018b931292330e519becd8d003ed0dd20c5e42de62ec3d37553698739aR16) `encryption_password_store TEXT NOT NULL CHECK (encryption_password_store IN ('user', 'mpc')),`
I propose to not to use `password` word here for `user` type of storage. I don't really like it, but I haven't found anything more suitable than `user` word.

PR #1009, opened 2025-07-31
feat(core): validate complex params with zod onm kwil actions

ditoglez:


packages/@core/tsconfig.json
```diff
@@ -16,6 +16,7 @@
     "composite": true,
     "declaration": true,
     "declarationMap": true,
+    "isolatedDeclarations": true,
```
Strnadj, line 19: why? I thought we want to get rid of those

Strnadj (approved):

PR #1133, opened 2025-09-15
feat(enclave): Encrypt encryptio sectet key & passwrod before storing…

Mohammed-Mamoun98:
… them using user-id


packages/utils/src/enclave/local.ts
```diff
@@ -88,6 +94,34 @@ export class LocalEnclave<
...
+    let password: string | undefined;
+    if (passwordBytes) {
+      password = new TextDecoder().decode(passwordBytes);
+    }
+
+    // Fallback to regular storage for migration compatibility
+    if (!password || !encryptionSecretKey) {
+      console.warn(
+        "⚠️ Falling back to unencrypted storage - consider migrating and deleting the old storage",
+      );
```
pkoch, line 114: Falling back is not an option. Refuse to operate unsecurely.

packages/utils/src/store/encryption-codec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  static createUserEncryptionCodec(userId: string): UserEncryptionCodec {
+    const userEncryptionKey = createUserEncryptionKey(userId);
+    return new UserEncryptionCodec(userEncryptionKey);
+  }
+
+  encode(data: Uint8Array): string {
+    try {
+      // Generate ephemeral key pair for this encryption
+      const ephemeralKeyPair = nacl.box.keyPair();
```
pkoch, line 47: How did you end up deciding on asymmetric encryption? `secretbox` with a constant nonce would work just fine for our use case.

packages/utils/src/enclave/local.ts
```diff
@@ -216,15 +250,22 @@ export class LocalEnclave<
...
+    // Store non-sensitive data in regular storage
     await this.store.set(STORAGE_KEYS.USER_ID, userId);
-    await this.store.set(STORAGE_KEYS.PASSWORD, password);
     await this.store.set(STORAGE_KEYS.ENCRYPTION_PASSWORD_STORE, encryptionPasswordStore);
-    await this.storeWithCodec.set(STORAGE_KEYS.ENCRYPTION_SECRET_KEY, keyPair.secretKey);
     await this.storeWithCodec.set(STORAGE_KEYS.ENCRYPTION_PUBLIC_KEY, keyPair.publicKey);
 
+    // Store sensitive data (password and secret key) in encrypted storage
+    // Convert password string to Uint8Array for encryption
+    const passwordBytes = new TextEncoder().encode(password);
```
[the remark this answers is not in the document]
Mohammed-Mamoun98: it just feels wierd to import more libs from idos-sdks because if you  checked the pkg.json we don't import alot of idos pkgs

PR #1136, opened 2025-09-16
feat(enclave): basic encryption on private materials

pkoch:
Stolen from #1133, so I can be the approved committer.


pkoch:
@ditoglez 🏓

packages/utils/src/enclave/local.ts
```diff
@@ -64,16 +74,83 @@ export class LocalEnclave<
...
+    const toKey = (userId: string, salt: Uint8Array) =>
+      syncScrypt(Utf8Codec.encode(userId), salt, 16384, 8, 1, nacl.secretbox.keyLength);
+
+    return {
+      encode(data: string): string {
+        if (!self.options.userId) {
+          throw new Error("User ID required for encryption");
+        }
+
+        const dataBytes = Utf8Codec.encode(data); // Convert string to Uint8Array
```
pkoch, line 94: This doesn't make sure sense to me. But, if it works, we can revisit it later.

pkoch:
Mo's changes look alright to me.

Issue #1127, opened 2025-09-10
IDOS-ENC-6: add a password strength indicator


pkoch:
Given the nature of the data, some user education is in order to avoid weak, very attackable, passwords.

pkoch:
Come to think of it, I'm ok with just acknowledging the problem, and fix it Soon™ (right after the release).

pkoch:
@Bbasche Ship delay until after launch?

PR #1150, opened 2025-09-22
fix(enclave): delete old keys every time we set the new ones

pkoch:
Follow-up to https://github.com/idos-network/idos-sdk-js/issues/1127


Mohammed-Mamoun98 (approved):
LGTM

PR #1209, opened 2025-10-01
feat: enclave and embed app improvements

ditoglez:


apps/idos-enclave/src/lib/enclave.ts
```diff
@@ -320,7 +320,7 @@ export class Enclave extends LocalEnclave<LocalEnclaveOptions> {
         }
 
         port1.close();
-        this.dialog?.close();
+        // this.dialog?.close();
```
Strnadj, line 323: are you sure about this?

Strnadj (approved):

PR #1211, opened 2025-10-02
Feature/add stellar to mpc

sapience:


.github/workflows/bundle-size.yml
```diff
@@ -23,15 +23,15 @@ jobs:
...
+          node-version-file: .node_version
 
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
         with:
           version: 10.15.0
 
       - name: Install dependencies
-        run: pnpm install --frozen-lockfile
+        run: rm -rf node_modules && pnpm install --force
```
pkoch: Very opposed to stop using `--frozen-lockfile`. I need to be convinced.

sapience: It was done to fix a build. For some reason (weird Vercel caching kinda) only this solved the issue.
I will try to get it back.

PR #1448, opened 2026-05-15
Add blob gateway credential storage, KGW-authenticated reads, and async Kwil tx polling

sapience:
## Summary

Migrates credential content handling from inline base64 blobs to **IPFS-backed blob gateway storage**, while keeping backward compatibility with legacy inline credentials. Adds KGW session support for authenticated blob reads on server-side flows, and replaces blocking Kwil `COMMIT` broadcasts with async broadcast + client-side `txInfo` polling.

### Blob gateway (`@idos-network/utils`)
- New `BlobGateway` client for upload/fetch against `/blob/v1/*`
- `createBlobContentReference()` builds `ipfs://` URIs with deterministic CID hashing
- `resolveCredentialEncryptedContent()` reads either legacy inline `content` or blob-backed `content_uri` / `content_size`
- Fetch hardening: response size cap, CID re-verification on read, clearer 401 vs 403 handling
- Fix browser `Illegal invocation` by binding `fetch` correctly

### KGW auth (`@idos-network/kwil-infra`)
- `createKgwAuthenticatedFetch()` — wraps fetch with NodeKwil KGW session cookie + retry on auth failure
- `createKgwAuthenticatedBlobGateway()` — blob gateway client wired to KGW session
- Upgrade to `@idos-network/kwil-js@0.1.4` for NodeKwil cookie helpers
- **`execute()` polling** — synchronous executes now broadcast async (`SYNC`) then poll `txInfo` with exponential backoff (300ms start, 1.5×, 2s cap, jitter, 90s deadline); exports `KwilTxFailedError` / `KwilTxPollTimeoutError`

### Kwil actions & validation
- `create_prelim_credentials_by_dwg`, `share_credential`, and related schemas updated for `content_uri` / `content_size`
- Zod validation for IPFS URI prefix and non-negative integer `content_size`
- Consolidated dual-mode read logic (fixes duplicated read paths)

### SDK surfaces
- **Client** (`@idos-network/client`, minor bump): optional `blobGatewayUrl` (falls back to `nodeUrl`); browser blob reads use `credentials: "include"`; new `getCredentialSharedContent()`; DWG/share flows upload blobs and store URIs
- **Issuer**: credential service uploads encrypted blobs before Kwil writes
- **Consumer**: uses `createKgwAuthenticatedBlobGateway`; `getCredentialSharedContentDecrypted()` resolves blob-backed shared credentials

### Docs & examples
- Updated client/issuer/consumer guides and READMEs
- Fixed apps/examples for new credential content model

## Test plan

- [ ] Issue a credential via DWG — verify blob upload + Kwil write with `content_uri` / `content_size`
- [ ] Decrypt credential on client (inline legacy + new blob-backed)
- [ ] Share credential — grantee consumer can fetch and decrypt via KGW-authenticated blob gateway
- [ ] Confirm browser client sends cookies on cross-origin blob reads
- [ ] Confirm server-side consumer/issuer reuse NodeKwil KGW session for blob fetches
- [ ] Verify Kwil execute succeeds under load (async broadcast + polling)
- [ ] Verify Kwil execute surfaces `KwilTxFailedError` on on-chain action failure
- [ ] Run package tests: `pnpm --filter @idos-network/utils test`, `pnpm --filter @idos-network/kwil-infra test`, `pnpm --filter @idos-network/consumer test`, `pnpm --filter @idos-network/issuer test`

## Release Notes

* **New Features**
  * Added blob gateway uploads for credential sharing/access grants and blob-backed encrypted credential content.
  * Enabled KGW-authenticated reads for both consumer and issuer credential flows.
  * Updated delegated credential sharing to use blob-backed encrypted content and issuer-signing-key–based copy references.

* **Bug Fixes**
  * Improved credential content decryption by resolving blob-backed encrypted bytes (including correct handling of `content_size`) with legacy inline-content fallback.

* **Documentation**
  * Updated guides and client quick starts for `blobGatewayUrl` and server-side `requestAccessGrant` using `issuerSigningSecretKey`.


packages/credentials/src/types/credentials.ts
```diff
@@ -4,14 +4,27 @@ export type idOSCredential = {
...
+  content_uri?: string | null;
+  content_size?: number | null;
   encryptor_public_key: string;
   issuer_auth_public_key: string;
   original_id?: string | null;
+  inserter_type?: string | null;
+  inserter_id?: string | null;
+};
+
+export type idOSCredential2 = Omit<idOSCredential, "content" | "content_uri" | "content_size"> & {
```
ditoglez: `idOSCredential2` sounds weird, just try to pronounce it 😊. I also see you omitting fields and then reintroducing them again but making them required.

packages/issuer/src/services/credential.service.ts
```diff
@@ -178,6 +254,74 @@ export class CredentialService {
...
+      dwg_grantee: delegatedWriteGrant.consumerWalletIdentifier,
+      dwg_issuer_public_key: delegatedWriteGrant.issuerPublicKey,
+      dwg_id: delegatedWriteGrant.id,
+      dwg_access_grant_timelock: delegatedWriteGrant.accessGrantTimelock,
+      dwg_not_before: delegatedWriteGrant.notUsableBefore,
+      dwg_not_after: delegatedWriteGrant.notUsableAfter,
+      dwg_signature: delegatedWriteGrant.signature,
+    };
+
+    console.log("create_preliminary_credentials_by_dwg payload");
```
ditoglez, line 170: Dont forget to cleanup logs

packages/issuer/package.json
```diff
@@ -38,7 +39,8 @@
   },
   "devDependencies": {
     "tiny-invariant": "catalog:core",
-    "tsdown": "catalog:build"
+    "tsdown": "catalog:build",
+    "vitest": "^4.1.5"
```
ditoglez: I would probably also move `vitest` to `catalogs`

packages/utils/src/blob-gateway/index.test.ts
```diff
@@ -0,0 +1,58 @@
+// @vitest-environment node
+
+import { describe, expect, it } from "vitest";
+
+import { BlobGateway, createBlobContentReference } from ".";
```
ditoglez: Avoid importing from `.`. Better to use a filename like `gateway.ts` and make `index.ts` export the needed  functionality.

packages/issuer/package.json
sapience: Good catch. Thanks!

packages/utils/src/blob-gateway/index.test.ts
sapience: Good catch. Thanks.

packages/client/src/index.ts
```diff
@@ -606,24 +638,54 @@ export class idOSClientLoggedIn implements Omit<Properties<idOSClientWithUserSig
...
       locked_until: lockedUntil,
       content_hash: contentHash,
     };
 
-    await this.shareCredential(insertableCredential);
+    await this.shareCredential(preliminaryCredential);
+    await this.blobGateway.uploadCredentialBlobs({
+      requestId: preliminaryCredential.request_id,
+      copy: content,
+    });
```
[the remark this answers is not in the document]
sapience: ThIs case is covered by cleanup mechanism implemented on nodes (schema) and blob gateway.

packages/issuer/src/services/credential.service.ts
```diff
@@ -173,9 +167,20 @@ export class CredentialService {
...
 
+    console.log("create_preliminary_credentials_by_dwg payload");
+    console.log(JSON.stringify(payload, null, 2));
+
     await createCredentialsByDwg(this.#kwilClient, payload);
+    await this.#blobGateway.uploadCredentialBlobs({
+      requestId,
+      original: originalCredential.encryptedContent,
+      copy: copyCredential.encryptedContent,
+    });
```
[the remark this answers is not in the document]
sapience: There is a failsafe mechanism on blob gateway side for this case. Alongside with cleanup mechanism.

docs/guide-client.md
```diff
@@ -118,17 +119,18 @@ The SDK decrypts through the enclave provider, so plaintext handling stays in th
...
 const grant = await idOSClientLoggedIn.requestAccessGrant("CREDENTIAL_ID", {
   consumerAuthPublicKey: "CONSUMER_AUTH_PUBLIC_KEY_HEX",
   consumerEncryptionPublicKey: "CONSUMER_ENCRYPTION_PUBLIC_KEY_BASE64",
+  issuerSigningKeyPair,
   lockedUntil: Math.floor(Date.now() / 1000) + 90 * 24 * 60 * 60, // optional
 });
 ```
 
-`requestAccessGrant(...)` creates a shared credential copy encrypted for the consumer and returns the created grant payload.
+`issuerSigningKeyPair` must match the credential's `issuer_auth_public_key`; Kwil verifies the copy's `content_uri` signature before accepting the preliminary share.
```
pkoch: We've came to the conclusion that this doesn't have to be true, but we want to force parity to simplify the system's mental model.

Am I remembering correctly, @sapience?

packages/credentials/src/types/credentials.ts
```diff
@@ -4,17 +4,31 @@ export type idOSCredential = {
   id: string;
   user_id: string;
   public_notes: string;
-  content: string;
+  content_uri: string;
+  content_size: number;
```
pkoch, line 8: This is telling me that we're going to need downtime to migrate all the credentials.

Is that where we landed? I'm not necessarily opposed, as long as we measure and it's not the end of the world.

packages/kwil-infra/src/create-kwil-client.ts
```diff
@@ -101,8 +107,18 @@ export class KwilActionClient {
...
       types: this.#actionTypes(params.name),
     };
-    const response = await this.client.execute(action, signer, synchronous);
-    return response.data?.tx_hash;
+    const response = await this.client.execute(action, signer, false);
+    const txHash = response.data?.tx_hash;
+
+    if (!txHash) {
+      return undefined;
+    }
```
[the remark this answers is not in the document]
pkoch: Maybe raise the error body content?

packages/credentials/src/types/credentials.ts
sapience: There is no downtime needed. We support both type of content storage for reading. So old credential will live in the DB until be migrated (in transactions) and new credentials will store their content in IPFS.

pkoch: That doesn't match what I see on https://github.com/idos-network/idos-schema/pull/121. I need to think about this again tomorrow with a fresh brain.

pkoch: Right, yep. The programmer needs to decide the sdk code version. 👍

sapience: I meant the newest sdk supports reading content of existing credentials from the DB and content of new created credentials from IPFS.

docs/guide-client.md
sapience: It has to be true. But naming can be confusing. An issuer is not an inserter/caller. So a user grants to an issuer (`issuer_key`) a DWG. Issuer has to issue a credential and sign the proof by the same `issuer_key`. 
At the moment we have a check
```sql
    $the_same_issuer := false;
    for $row in SELECT 1 FROM delegates d1 INNER JOIN delegates d2 ON d1.inserter_id = d2.inserter_id
        WHERE d1.address = lower($issuer_auth_public_key) AND d2.address = lower($dwg_issuer_public_key) LIMIT 1 {
        $the_same_issuer := true;
        break;
    }
    if !$the_same_issuer {
        error('credentials issuer must be an issuer of delegated write grant');
    }
 ```
which we have to change to simple 
```sql
    if $issuer_auth_public_key  !=$dwg_issuer_public_key {
        error('credentials issuer must be an issuer of delegated write grant');
    }
 ```
 So the solution is to introduce a new `issuer_key` in SDK and distinct it from `delegate_key`.

File: packages/client/src/index.ts [hunk 10]
```diff
@@ -136,7 +150,13 @@ export class idOSClientIdle {
 
     await params.enclaveProvider.load();
 
-    return new idOSClientIdle(params.store, kwilClient, params.enclaveProvider);
+    const blobGateway = new BlobGateway({
+      url: params.blobGatewayUrl ?? params.nodeUrl,
+      // Kwil RPC uses withCredentials; blob reads need the same KGW session cookie cross-origin.
+      fetchFn: (input, init) => fetch(input, { ...init, credentials: "include" }),
```
pkoch: Just realized that we're indeed married to having the blob gateway be on the same host as KGW. Otherwise, they can't share `HttpOnly` cookies.

Not a bad call, it's one less way to get hacked. But I wanted to point it out specifically.

packages/client/src/index.ts
```diff
@@ -120,12 +127,19 @@ export class idOSClientIdle {
   readonly store: Store;
   readonly kwilClient: KwilActionClient;
   readonly enclaveProvider: BaseProvider;
+  readonly blobGateway?: BlobGateway;
```
pkoch: Why would we ever allow not having a BlobGateway client?

From what I could trace with a quick skim, that should never happen. Adding optionality takes crispness away from the types, and makes future maintainers double-guess why this was made optional to begin with.

packages/client/src/index.ts
```diff
@@ -563,7 +584,9 @@ export class idOSClientLoggedIn implements Omit<Properties<idOSClientWithUserSig
       );
 
       const matches = await this.enclaveProvider.filterCredentials(
-        fullCredentials,
+        await Promise.all(
+          fullCredentials.map((credential) => this.#credentialWithInlineContent(credential)),
+        ),
```
pkoch, line 591: Come to think of it, the concept of a full credential is losing its value. It came into existence for us to not have to fetch a gazzilion bytes for each row. But, since we're now moving those bytes elsewhere, that point is moot.

Not something to action now, but something to keep in mind. 👁️ @Strnadj

packages/kwil-infra/src/create-kgw-authenticated-fetch.ts
```diff
@@ -0,0 +1,259 @@
...
+  const nodeKwil = getNodeKwilClient(kwilClient);
+  let refreshPromise: Promise<string> | undefined;
+
+  const refreshCookie = async (): Promise<string> => {
+    refreshPromise ??= nodeKwil.authenticateKGWAndSetCookie(signer).finally(() => {
+      refreshPromise = undefined;
+    });
+
+    return refreshPromise;
+  };
```
pkoch, line 60: This one took me a good while to build a mental model for. We should invest in making this simpler for the next reader, or explain why it can't be simpler.

Why are we babysitting a singleton? I can't figure out a race condition from reading the code, since we always await both `ensureCookie` and `refreshCookie`.

packages/kwil-infra/src/create-kgw-authenticated-fetch.ts
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  if (!hasKgwSessionHelpers(kwilClient.client)) {
+    throw new Error("NodeKwil client does not expose KGW session cookie helpers");
+  }
+
+  return kwilClient.client;
+}
+
+function hasKgwSessionHelpers(client: NodeKwil): client is KgwSessionNodeKwil {
```
pkoch, line 265: Mini nit: I'd love to start all my type narrowers with `is`.

packages/kwil-infra/src/wait-for-kwil-tx.ts
```diff
@@ -0,0 +1,110 @@
...
+  return txInfo.height > 0;
+}
+
+export async function waitForKwilTx(
+  txInfo: (txHash: string) => Promise<{ data?: KwilTxInfo }>,
+  txHash: string,
+  deadlineMs: number = KWIL_TX_POLL_DEADLINE_MS,
+): Promise<void> {
+  const deadlineAt = Date.now() + deadlineMs;
+  let intervalMs = INITIAL_POLL_INTERVAL_MS;
```
pkoch, line 73: I dread the day we have to distinguish miliseconds from megaseconds. :P

packages/kwil-infra/src/wait-for-kwil-tx.ts
```diff
@@ -0,0 +1,110 @@
...
+      const response = await txInfo(txHash);
+      const data = response.data;
+
+      if (data && isTxFinalized(data)) {
+        if (data.tx_result.code === 0) {
+          return;
+        }
+
+        throw new KwilTxFailedError(txHash, data.tx_result.code, data.tx_result.log, data.height);
+      }
```
pkoch: Fallthrough logic is not the easiest to read. Maybe something like:
```suggestion

      if (!data) { throw new MalformedRespose("etc"); }
      if (isTxFinalized(data)) {
        if (data.tx_result.code === 0) { return; }

        throw new KwilTxFailedError(txHash, data.tx_result.code, data.tx_result.log, data.height);
      }
      lastPollError = new TxStillNotFinalized("etc")
```

Took me a little while to decode what it meant for:
- the if to not be triggered
- `lastPollError` to be `undefined`

packages/utils/src/blob-gateway/index.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    });
+
+    const result = await gateway.uploadCredentialBlobs({
+      requestId: "request-1",
+      original: utf8Encode("original"),
+      copy: utf8Encode("copy"),
+    });
+
+    expect(result).toEqual({ request_id: "request-1", quorum: 1 });
+    expect(calls[0]?.[0]).toBe("https://blob.example/blob/v1/requests/request-1/upload");
```
pkoch, line 98: Is this using implicit string conversion? Can we maybe make it explicit?

pkoch: No, nvm, misread.

packages/utils/src/blob-gateway/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      throw error;
+    }
+
+    const responseText = await response.text();
+
+    if (!response.ok) {
+      throw new Error(`blob gateway upload failed with ${response.status}: ${responseText}`);
+    }
+
+    return JSON.parse(responseText) as BlobGatewayUploadResponse;
```
pkoch: I almost feel tempted to user Zod here. Almost.

packages/utils/src/blob-gateway/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      throw new Error(
+        `blob gateway expected size ${expectedByteLength} exceeds maximum fetch size ${maxFetchBytes}`,
+      );
+    }
+
+    const maxResponseBytes = expectedByteLength ?? maxFetchBytes;
+    const fetchUrl = `${this.#url}/blob/v1/ipfs/${encodeURIComponent(cid)}`;
+
+    const response = await this.#fetch(fetchUrl);
+    const contentLength = response.headers.get("content-length");
```
pkoch, line 195: Keep in mind that not returning a `content-length` header is not that unusual.

packages/kwil-infra/src/actions/actions.test.ts
```diff
@@ -0,0 +1,126 @@
...
+};
+
+describe("CreateCredentialsByDwgInputSchema", () => {
+  it("accepts IPFS content URIs and positive integer sizes", () => {
+    expect(() =>
+      CreateCredentialsByDwgInputSchema.parse(validPreliminaryCredentialsByDwgInput),
+    ).not.toThrow();
+  });
+
+  it("rejects non-IPFS content URIs", () => {
```
pkoch: `it.each`?

packages/kwil-infra/src/create-kgw-authenticated-fetch.ts
pkoch: `it("shares one refresh across parallel auth failures", async () => {`. I see.

Since it took me a while to get it, I'd add a comment for the next reader? Unsure if it was a me problem because I just had too much tunnel vision on "node is not threaded" when reading.

docs/guide-client.md
pkoch: Makes sense. Turning on gas might make us realize we don't need inserters anymore, even.

pkoch (approved):
Bunch of nits, but nothing egregious.

Ready enough for prime time. 🚀

examples/consumer-and-issuer/next-env.d.ts
```diff
@@ -0,0 +1,5 @@
+/// <reference types="next" />
+/// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
```
Strnadj: what is this?

docs/guide-consumer.md
```diff
@@ -336,13 +336,7 @@ This will return a list of `idOSCredentials` that match the filtering criteria.
...
 ### [ frontend ] Requesting access grant
 
-The simplest way to do this is to ask the user to create and insert an access grant for you.
-
-```typescript
-const accessGrant: idOSGrant = await idOSClient.requestAccessGrant("CREDENTIAL_ID");
-```
-
-Alternatively, you can ask for a delegated access grant, which the user creates:
+Ask the user to create a delegated access grant, then insert it from your backend:
```
Strnadj: I am missing something here?

examples/pay-demo/app/providers/state/flows/due.ts
```diff
@@ -232,7 +232,7 @@ export const actors = {
         "Content-Type": "application/json",
       },
       body: JSON.stringify({
-        dagId: input.ag?.id,
+        dagId: input.ag?.copy_id,
```
Strnadj, line 235: Why did you change this? 🤔

packages/client/src/index.ts
Strnadj: That means that the `blobGatewayUrl` should disappear?

packages/client/src/index.ts
Strnadj: Let's make it mandatory.

packages/consumer/src/index.ts
```diff
@@ -149,6 +162,10 @@ export class idOSConsumer {
   ): Promise<VerifyCredentialResult> {
     return verifyCredential<K>(credentials, issuers);
   }
+
+  async #getCredentialEncryptedContent(credential: idOSCredential): Promise<string> {
+    return base64Encode(await resolveCredentialEncryptedContent(credential, this.#blobGateway));
```
Strnadj: I am not super sure about this, but we can essentially get rid of `base64Encode` because what happens...

We have a Uint8Array (or Buffer) in Node, which has been obtained from the blob gateway. Then we do the base64Encode and almost immediatelly we are sending those to `noncedBox.decrypt` which leads to nacl.box.open that requires again Uint8Array / Buffer.

Let's remove the encoding in the way we are dealing with several megabytes of data. Those are making things slower. 

(cc @pkoch )

packages/utils/src/blob-gateway/index.ts
```diff
@@ -0,0 +1,304 @@
+import { of as ipfsOnlyHash } from "ipfs-only-hash";
+
+import { base64Decode } from "../codecs";
+
+export type BlobContentReference = {
```
Strnadj, line 5: This whole thing feels more like a `kwil-infra` kinda thing, not a utility. I would rather move it there.

packages/kwil-infra/src/actions/actions.ts
```diff
@@ -18,6 +18,10 @@ export const encryptionPasswordStoreSchema: z.ZodType<EncryptionPasswordStore> =
   ENCRYPTION_PASSWORD_STORES,
 );
 
+const IPFS_URI_PREFIX = "ipfs://";
```
Strnadj, line 20: This whole file is generated from a schema, but I can't see any changes that will produce this 3 lines... 🤔

packages/client/src/index.ts
sapience: > That means that the `blobGatewayUrl` should disappear?

No. It can help to setup local development env.

sapience: I added more comments regarding it.

packages/kwil-infra/src/create-kgw-authenticated-fetch.ts
sapience: Added the comment.

packages/kwil-infra/src/wait-for-kwil-tx.ts
sapience: Refactored

packages/kwil-infra/src/actions/actions.test.ts
sapience: Refactored.

examples/consumer-and-issuer/next-env.d.ts
sapience: This is an auto-generated TypeScript declaration file that Next.js creates/updates for the consumer-and-issuer  project.

packages/client/src/index.ts
sapience: I can say from other angle: concept of shared credential is loosing its value. It seems we need only one (original) credential and AGs for it as a shared material.

docs/guide-consumer.md
sapience: I missed :) I forgot to delete DAG usage from everywhere.
It was removed in this [commit](https://github.com/idos-network/idos-sdk-js/pull/1448/changes/64307c089e24175d91c495123bc37de52c1daf89). Please take a look.

examples/pay-demo/app/providers/state/flows/due.ts
sapience: Changed in 42031293 (“Fix apps and examples”) because relayAG is the return value of requestAccessGrant(), and that object has no .id.

It returns a ShareCredentialInput shaped like:

copy_id — shared credential ID (what the AG is for)
original_id — user’s original credential
request_id, etc.
So input.ag?.id was always undefined. The Due KYC route needs the shared copy ID for fetchSharedToken(...), hence copy_id. Same fix was applied in actors.ts for the token URL.

packages/consumer/src/index.ts
sapience: Good catch! Round-trip was removed.

packages/kwil-infra/src/actions/actions.ts
sapience: You are right. I [updated](https://github.com/idos-network/idos-schema/pull/121/commits/7f4b37cb0e3b7d53da51a87c1ce039f4ee7a8513) the generator and regenerate the file.

packages/kwil-infra/src/create-kgw-authenticated-fetch.ts
```diff
@@ -0,0 +1,263 @@
...
+  };
+
+  return async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+    const { cookie } = await ensureCookie();
+    const bufferedRequest = await bufferRequestBody(input, init);
+    const response = await fetchFn(
+      ...withCookie(input, withMaterializedBody(bufferedRequest), cookie),
+    );
+
+    if (!(await isAuthFailure(response.clone()))) {
```
Strnadj: Why is there a clone? I am not sure how big those responses could be.

Strnadj: Also the "isAuthFailure" calls `.json` are you sure that you can do that multiple times? 🤔

packages/utils/src/blob-gateway/index.ts
Strnadj: AI has a point in here... ```Upload response CIDs are never checked. BlobGatewayUploadResponse carries original_cid / copy_cid / quorum; nothing compares them to the locally computed CIDs or asserts quorum. Compare and throw — otherwise a gateway that stored something else is only discovered later, on read, as a mismatch.```

packages/issuer/src/services/credential.service.ts
```diff
@@ -173,9 +167,19 @@ export class CredentialService {
       dwg_signature: delegatedWriteGrant.signature,
     };
 
-    await createCredentialsByDwg(this.#kwilClient, payload);
+    console.log("create_preliminary_credentials_by_dwg payload");
+
+    await createPreliminaryCredentialsByDwg(this.#kwilClient, payload);
```
Strnadj, line 172: This is not "just" a 1 request to Kwill to insert credentials, what's gonna to happen if the upload fails?

Strnadj: (is there any rollback, another upload attempt, etc?)

packages/kwil-infra/src/create-kgw-authenticated-fetch.ts
sapience: Mostly yes — the instinct is right, with one nuance on when size matters.

Why clone?
A Response body is a one-shot stream. defaultKgwAuthFailurePredicate may call response.json(), which consumes the body. Without a clone, a successful (non-auth-failure) response would be returned already drained, and the caller couldn’t read it. Clone gives the predicate its own body; the original stays readable for the return path.

Can you .json() twice?
Not on the same Response. You can .json() the clone and later read the original — that’s the whole point of clone().

Size
clone() does not copy the whole body into memory up front; it tees the stream. The real risk is the unread branch: for a large blob fetch, the predicate often returns early on content-type (no read), then you return the original and leave the clone hanging. Tee’d data can still buffer on that unused side.

So: clone is correct for the one-shot body rule; worrying about large responses is fair for the abandoned clone, not because .json() runs twice. For blob downloads, cancelling the unused clone when it’s not an auth failure (e.g. clone.body?.cancel()) would address that.

sapience: Fix: clone only when content-type is JSON (the only case the default predicate reads the body). Blob downloads skip the clone entirely. On auth failure, cancel the discarded response so it doesn’t keep streaming.

examples/consumer-and-issuer/next-env.d.ts
Strnadj: but there is no consumer-and-issuer project anymore...

sapience: I haven't notice that. Thank you!

packages/utils/src/blob-gateway/index.ts
sapience: Done.

packages/issuer/src/services/credential.service.ts
sapience: No changes here. The behavior is the same as was in `just a 1 request to Kwill to insert credentials`. No rollback (nothing to rollback), no another upload attempt. And i see no reasons to add them.
If the Kwil action fails, then UX will show it the user has to click something (a lot of reason why it can be).
What case do you have in mind? And was that case handled earlier (before IPFS)?

packages/issuer/src/services/credential.service.ts
sapience: Thank you!

packages/issuer/src/services/credential.service.ts
Strnadj: @pkoch would you mind stepping in, maybe I  don't understand it at all... :-( 

Previously it was:
```
1. encrypt content
2. use DWG to write credentials (and the copy for DAG)
3. done
```

Now it's:
```
1. encrypt content
2. use DWG to write credentials metadata
3. use blob GW to upload credentials
4. done
```

I would expect some kind of transaction or rollback for `2` and `3` step.

pkoch: There is. Look for the preliminary DWG stuff on the schema.

PR #1482, opened 2026-07-24
Add mm_token Kwil signer and MM wallet type support

sapience:
## Summary
- Add `createMmTokenKwilSigner` so clients can authenticate to KGW with `mm_token` envelopes without changing normal `call` / `execute` usage
- Add unpadded base64url codec helpers used by the envelope/token decoding path
- Thread `MM` through wallet type schemas and treat trusted-inserter MM wallets as valid without local signature verification

## Details
- Signer identity is the decoded `signing_public_key` bytes (`Uint8Array`), not the base64url string
- `signMessage` returns the exact envelope bytes KGW expects as signature data
- `upsertWalletAsInserter` can accept `wallet_type: "MM"` with `address` and `public_key` set to the same base64url signing key

## Test plan
- [ ] `pnpm --filter @idos-network/utils test -- --run src/codecs/index.test.ts`
- [ ] `pnpm --filter @idos-network/kwil-infra test -- --run src/mm-token/create-mm-token-kwil-signer.test.ts src/signature-verification/index.test.ts src/actions/actions.test.ts`
- [ ] Build an envelope, call `createMmTokenKwilSigner(...)`, set it on a Kwil client, and run a read + write action
- [ ] Confirm trusted inserter can upsert an `MM` wallet with matching `address` / `public_key`
- [ ] Confirm invalid token base64 / wrong key length fail with clear errors

## Summary by CodeRabbit

- **New Features**
  - Added MetaMask token authentication for MM wallets, including UKYC credential references and access-token authorization.
  - Added Node.js client support with in-memory storage and runtime-specific blob gateways.
  - Added encrypted credential creation, retrieval, deletion, and rescission workflows.
  - Added support for UKYC content URIs alongside IPFS references.
  - Added URL-safe Base64 encoding and decoding utilities.

- **Bug Fixes**
  - Improved profile and credential validation for MM encryption stores and wallet types.
  - Added clearer handling for missing users, invalid selections, and unauthorized UKYC access.

- **Documentation**
  - Added MM token authentication and wallet-management guidance.


packages/kwil-infra/src/signature-verification/index.ts
```diff
@@ -18,9 +18,15 @@ export interface WalletSignature {
 
 export const verifySignature = async (walletPayload: WalletSignature): Promise<boolean> => {
   invariant(walletPayload.address, "Wallet address is required");
+  invariant(walletPayload.wallet_type, "Wallet type is required");
+
+  if (walletPayload.wallet_type === "MM") {
```
Strnadj, line 23: This is duplicates of line 25...

packages/client/src/index.test.ts
```diff
@@ -0,0 +1,281 @@
...
+  };
+}
+
+describe("credential blob storage", () => {
+  it("creates MM credentials with ukyc:// and the encrypted byte size", async () => {
+    const { credential, preliminaryInput, uploadedBytes } = await createCredentialForWallet("MM");
+
+    expect(credential.content_uri).toMatch(/^ukyc:\/\/storage-abc\/blobs\//);
+    expect(preliminaryInput?.content_uri).toBe(credential.content_uri);
+    expect(preliminaryInput?.content_size).toBe(credentialPlaintext.byteLength + 40);
```
pkoch, line 269: Where does the 40 come from?

packages/client/src/index.test.ts
```diff
@@ -0,0 +1,281 @@
...
+  } as unknown as idOSClientWithUserSigner;
+  const client = new idOSClientLoggedIn(withSigner, {
+    id: crypto.randomUUID(),
+    recipient_encryption_public_key: base64Encode(new Uint8Array(32).fill(7)),
+    encryption_password_store: "user",
+  });
+
+  return {
+    credential: await client.createCredential("public notes", credentialPlaintext),
+    preliminaryInput,
```
pkoch, line 258: I would assert this is present to make the function signature easier to work with.

packages/client/src/index.ts
```diff
@@ -630,13 +763,18 @@ export class idOSClientLoggedIn implements Omit<Properties<idOSClientWithUserSig
 
     // User issues the shared copy: sign content_uri with an ephemeral user-side key.
     // Copy issuer_auth_public_key need not match the original credential's issuer.
-    const copyReference = await createBlobContentReference(content);
-    const signedReference = buildEphemeralSignedCredentialContentReference("", copyReference.uri);
+    const copyId = crypto.randomUUID();
+    const ukycContentUri = contentUriForWallet(this.walletType, this.kwilSigner, copyId);
+    requireAccessTokenForUkycContent(ukycContentUri, this.blobGateway.hasAccessToken);
```
pkoch, line 768: Something about using `requireAccessTokenForUkycContent` so often smells. Can't quite put my finger on it yet, tho.

packages/client/src/index.ts
```diff
@@ -314,15 +377,79 @@ export class idOSClientLoggedIn implements Omit<Properties<idOSClientWithUserSig
...
 
   async requestDWGMessage(params: idOSDelegatedWriteGrant): Promise<string> {
     return dwgMessage(this.kwilClient, params).then((res) => res.message);
   }
 
+  async createCredential(
+    publicNotes: string,
+    plaintextContent: Uint8Array<ArrayBufferLike>,
+    // In case of credential creation on users behalf, this would be a issuer key
+    issuerSigningSecretKey?: Uint8Array,
```
pkoch, line 396: I think it make sense to future proof this just a tiny bit. Instead of passing raw secret material, pass a signer instead. This way, we could even automate filling in a column that describes how things are signed/encrypted.

I'd not block on this. Give this task to an AI friend and see if you can get this for almost free. If not, I'm super ok to drop.

packages/consumer/src/index.ts
```diff
@@ -111,7 +122,25 @@ export class idOSConsumer {
...
 
   async rescindSharedCredential(credentialId: string): Promise<void> {
-    return rescindSharedCredential(this.#kwilClient, { credential_id: credentialId });
+    const credential = await this.getCredentialSharedFromIDOS(credentialId);
+    invariant(credential, `Credential with id ${credentialId} not found`);
+
+    const contentUri = credential.content_uri ?? null;
+    requireAccessTokenForUkycContent(contentUri, this.#blobGateway?.hasAccessToken ?? false);
+
+    await rescindSharedCredential(this.#kwilClient, { credential_id: credentialId });
```
pkoch, line 131: I'm just realizing we can leave garbage behind. I vaguely remember having a "list all URIs that the DB refers to, and delete the others". If we didn't do that, this can leave garbage in the blobl storage (because we're not sure if `this.#blobGateway.deleteCredentialBlob({ credentialId });` will ever run).

packages/kwil-infra/src/signature-verification/index.ts
pkoch: Indeed. `walletPayload.wallet_type === "MM"` implies `walletPayload.wallet_type !== "EVM"`

packages/consumer/src/index.ts
pkoch: I was looking at the wrong version of the schema. We do mark this for GC:
```sql
    INSERT INTO credential_deletion_requests (
        credential_id,
        content_uri,
        content_size,
        requested_by_wallet_identifier,
        created_at
    ) VALUES (
        $credential_id,
        $content_uri,
        $content_size,
        @caller,
        @block_timestamp
    );
```

packages/credentials/src/utils/index.ts
```diff
@@ -184,14 +185,73 @@ export function buildSignedCredentialContentReference(
...
+}
+
+export async function buildPreliminaryIDOSCredential({
+  publicNotes,
+  plaintextContent,
+  recipientEncryptionPublicKey,
+  // For user-issued credentials, use a fresh ephemeral key for the signed reference
+  // For issuer-side, use the signing key pair
+  issuerSigningSecretKey = nacl.sign.keyPair().secretKey,
+  contentUri,
```
pkoch, line 229: A mandatory param after an optional param is a recipe for disaster. :/

pkoch: Oh, this is destructuring an object, nevermind me! Resolving.

packages/kwil-infra/src/signature-verification/index.test.ts
```diff
@@ -107,4 +107,22 @@ describe("verifySignature", () => {
...
+      wallet_type: "MM",
+      public_key: ["mm-signing-public-key"],
+    });
+
+    expect(result).toBe(true);
+    expect(verifyEvmSignature).not.toHaveBeenCalled();
+    expect(verifyNearSignature).not.toHaveBeenCalled();
+    expect(verifyRippleSignature).not.toHaveBeenCalled();
+    expect(verifyStellarSignature).not.toHaveBeenCalled();
+  });
```
pkoch, line 127: I'm going to need a refresher on this one.

pkoch (commented):
I need help confirming my understanding on a couple of things, but looks good overall.

packages/client/src/index.ts
```diff
@@ -314,15 +377,79 @@ export class idOSClientLoggedIn implements Omit<Properties<idOSClientWithUserSig
...
+    const contentUri = credential.content_uri ?? null;
+    requireAccessTokenForUkycContent(contentUri, this.blobGateway.hasAccessToken);
+
     await removeCredential(this.kwilClient, { id });
+
+    if (!contentUri) {
+      return { id };
+    }
+
+    await this.blobGateway.deleteCredentialBlob({ credentialId: id });
```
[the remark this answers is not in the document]
sapience: The deletion has fail safe mechanism. Blob-gateway will handle the case.

packages/consumer/src/index.ts
```diff
@@ -111,7 +122,25 @@ export class idOSConsumer {
...
+      return;
+    }
+
+    if (!this.#blobGateway) {
+      throw new Error(
+        `Credential with id ${credentialId} is blob-backed, but blobGatewayUrl was not configured`,
+      );
+    }
+
+    await this.#blobGateway.deleteCredentialBlob({ credentialId });
```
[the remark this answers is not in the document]
sapience: Blob-gateway handles such deletion cases.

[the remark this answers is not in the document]
sapience: I meant blob-gateway as a separate service. Yes i confirm.

packages/client/src/index.ts
sapience: The suggestion is reasonable but not “almost free” here.

Currently the credentials utility itself requires an Ed25519 secret key and derives the public key from it. Passing a signer only at createCredential() would therefore just move the raw-key conversion elsewhere.

A proper signer abstraction needs:

```js
type CredentialSigner = {
  publicKey: Uint8Array;
  sign(message: Uint8Array): Promise<Uint8Array>;
  algorithm: "Ed25519";
};
```

That would affect the credentials builder, issuer service, client API, examples, and tests. It also requires deciding how the algorithm metadata maps to the proposed database column.

I’d defer this non-blocking suggestion into a dedicated refactor rather than introduce a partial abstraction. No code changed.

pkoch: uff, yes. I was expecting online the sign method. Nevermind me, opportunity for next time!

pkoch (approved):

PR #1500, opened 2026-08-14
feat(client) Expose create credentials method

Strnadj:


sapience (changes requested):
# Review: `feature/expose-create-credentials-for-issuer`

Compared against `origin/feature/mm_auth_check`.

## Recommendation

Do not merge this branch as-is, but the main motivation is valid.

If the goal is to create credentials using MM-token authorization, then exposing a helper around `create_preliminary_credential` makes sense. This can reasonably serve both intended cases:

- a user creating a credential for themselves while authenticated with their MM token;
- an issuer/relay creating a credential on the user's behalf while presenting a user-authorized MM token.

In both cases, KGW resolves `@caller` from the MM token's `signing_public_key`, so schema authorization is user-scoped.

Credential provenance is separate. If an issuer signing key is passed, `issuer_auth_public_key` should reflect that issuer. If no issuer signing key is passed, the method should clearly document that it uses an ephemeral/user-side signing key and does not prove issuer provenance.

## Findings

### Wrong default signing key type

`buildPreliminaryIDOSCredential()` defaults `signedReferenceSecretKey` to `nacl.box.keyPair().secretKey`, but the value is passed into `buildSignedCredentialContentReference()`, which uses `nacl.sign.keyPair.fromSecretKey()`.

That default should be an Ed25519 signing key:

```ts
nacl.sign.keyPair().secretKey
```

or the signing key should be required.

For the MM issuer flow, requiring the issuer signing key may be better than silently generating an ephemeral one, because credential provenance should come from the issuer.

### Inconsistent content hash semantics

`idOSClientLoggedIn.createCredential()` sets `content_hash` from the encrypted blob bytes:

```ts
hexEncodeSha256Hash(preliminaryCredential.encryptedContent)
```

Existing SDK flows treat `content_hash` as the hash of plaintext credential content:

- `getCredentialContentSha256Hash()` decrypts and hashes plaintext.
- `requestAccessGrant()` hashes plaintext before sharing.
- issuer DWG creation hashes `credentialParams.plaintextContent`.

The new method should use:

```ts
hexEncodeSha256Hash(plaintextContent)
```

### API should make the MM-token identity split clear

The helper can support both intended MM-token flows:

- user self-create with their own MM token;
- issuer/relay create on the user's behalf with a user-authorized MM token.

In both cases, KGW authorization is user-scoped, because `@caller` comes from the MM token.

Issuer provenance is separate from authorization. The API should document whether `signedReferenceSecretKey` is:

- required for issuer-created credentials;
- optional for user-created credentials;
- or optional everywhere, with ephemeral signing used when omitted.

The current shape can work, but this behavior should be explicit because `issuer_auth_public_key` is part of the credential metadata users and consumers may rely on.

## Changes Worth Keeping

- Extracting shared preliminary credential-building code into `@idos-network/credentials/utils` is useful if both DWG and MM-token creation flows need it.
- Adding a high-level helper for `create_preliminary_credential` is useful for the MM-token flow.
- The enclave guard that rejects `"mm"` as an auth method is reasonable after widening `EncryptionPasswordStore`, even though the enclave does not implement MM.

## Before Merge

1. Fix the default signing key type or require a signing key.
2. Store plaintext content hash in `createCredential()`.
3. Document the two-identity MM flow: user authorization via MM token, issuer provenance via issuer signature.
4. Add tests for:
   - `buildPreliminaryIDOSCredential()` with an issuer signing key;
   - the default/no-key behavior, if kept;
   - the generated `create_preliminary_credential` payload;
   - plaintext `content_hash`.

## Test Note

Attempted to run the credentials package tests, but pnpm aborted because it wanted to purge `node_modules` in a non-TTY install step. No test result was collected.

